PR #355, opened 2015-01-13
@orta => Cleanup

ashfurrow:
This cleans up a lot of things, particularly surrounding pods that are now on trunk. I opted to write our own logger because we only need it to be very simple – all of the existing solutions were overkill imo.


orta:
This makes sense. Mergable on :green_apple:

ashfurrow:
:+1:

PR #370, opened 2015-02-11
@orta => Added BuyersPremium

### LINKED ISSUE #360: Buyer's Premium
Waiting on API to be updated and some designs from @katarinabatina.

ashfurrow:
There are specifics attached to the "schedule" in the JSON from staging – not sure if we're using this yet. ~~Waiting to hear back from the auctions team if the minimum bid is being used and would need to be implemented as well.~~

Tasks:

- [x] Add/parse model
- [x] Augment UI/make interactive
- [x] Tests

Fixes #360.


orta:
This looks great to me, little bit un-DRY with the long repetition around the buyers premium label, perhaps a UIViewSubclass of its own?

PR #379, opened 2015-02-18
@alloy => No longer render markdown links

ashfurrow:
This is a WIP until @xing can take a look at https://github.com/xing/XNGMarkdownParser/pull/5 

Fixes #375.

PR #381, opened 2015-02-19
[Exploration] Build before test

pietbrauer:
I found out that in the xcodebuild commands it is required to have `build test` instead of `test` for Xcode to launch the simulator.
Otherwise, if your build is taking longer than 2 minutes, it aborts because of the timeout.
As I saw in your Makefile your are just calling `test` so that might fix the problem described in https://github.com/artsy/eidolon/pull/379#issuecomment-74949058?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ clean:
 	xcodebuild -workspace '$(WORKSPACE)' -scheme '$(SCHEME)' -configuration '$(CONFIGURATION)' clean
 
 test:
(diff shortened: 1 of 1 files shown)


ashfurrow:
Cool. Failure on CI is due to lack of access to Travis secure environment variables on PRs from forks. Meeerging thanks!

alloy:
@pietbrauer Such spotting, amaze! :heart:

pietbrauer:
Did it work?

PR #386, opened 2015-02-27
@alloy => Makes things more ReactiveCocoa-y

### LINKED ISSUE #297: Bid numbers not deleting
Saw a user at #CMAArtAuction placing a bid, enter a price, then try to delete the value. None of the delete/clear buttons were working. They canceled out and restarted and everything worked.

ashfurrow:
Initial work on RAC'ifying things. I don't think it even compiles now, but wanted to get something in before EOD. 

Basically what this does is take a whole bunch of disparate almost-identical code around the app and puts it in one place. This forms an abstraction around the `keypadVIew` in the form of the new view model, and abstracts _that_ away so you only deal with the `keypadContainerView` because the Law of Demeter or whoever. 

Tested manually and it's :100: 

TODO: 
- [x] Needs tests. 
- [x] The three view controllers are maintaining instance copies of state and I don't think this is necessary. 
- [ ] I'd like to get @orta's thoughts sometime (not necessary for merging, but :+1:).

Fixes #297. Probably.


ashfurrow:
Ha, yeah. Compiler is segfaulting.

ashfurrow:
Seems like it's crashing on the newly added file, `KeypadViewModel.swift`. 

```
Global is external, but doesn't have external or weak linkage!
i8* ()* @_TFFC5Kiosk15KeypadViewModelg15addDigitCommandCSo10RACCommandaL_4selfS0_
invalid linkage type for function declaration
i8* ()* @_TFFC5Kiosk15KeypadViewModelg15addDigitCommandCSo10RACCommandaL_4selfS0_
LLVM ERROR: Broken module found, compilation aborted!
```

ashfurrow:
Looks similar to [this problem](http://stackoverflow.com/questions/25430102/linkage-error-when-calling-super-from-property-override) involving accessing `super` from a  property override.

alloy:
Fun times :)

ashfurrow:
Realized there was a problem with this as it was – phone numbers and PINs/bidder numbers that started with zero would have that prefix chopped off. Adding tests now.

ashfurrow:
This is mergeable. Same thing as in eigen – merging on reds until Travis gets their house in order, or until we have time to replace them.

ashfurrow:
Hmm. Probably could use `waitUntil` syntax. I'll throw it in tomorrow, maybe in another PR.

alloy:
The CI is green right now, so I’ll merge this and you can do the ‘wait’ stuff in another PR.

ashfurrow:
Changes made in: https://github.com/artsy/eidolon/pull/392

PR #407, opened 2015-03-27
@sarahscott => Fix tests

### LINKED ISSUE #405: Regenerate snapshots
Once #404 is merged, it will start legitimately failing on Travis due to problems with the snapshots. I can't generate new ones based on the changes because I don't have the 8.1 simulator installed and can't install it right now. 

Use 8.2 if Travis updates or if we move to another CI provider by then.

ashfurrow:
Specifically, updates an outdated snapshot and adds better use of DI in the listings VC tests.

This likely won't be :green_apple: on Travis because we're having Travis issues. 

Fixes #405.


ashfurrow:
Whaaaaat a legit Travis failure? Looking into it now.

.travis.yml
```diff
@@ -12,7 +12,7 @@ install:
   - make prepare_ci
   - make bundle
 before_script:
-  - make ci
+  - travis_retry make ci
```
ashfurrow: "Get your builds to compile with this one weird trick!"

orta: o.0

orta: read up, smart.

PR #408, opened 2015-03-30
@orta => Fixes many things

### LINKED ISSUE #397: Lots of println
Using `println` avoids our logging library, which could cause problems when trying to troubleshoot problems at an auction.

ashfurrow:
## Removes unnecessary view controllers from storyboard

If you looked at the storyboard, there were two other VCs containing the starting VC. 

screen shot 2015-03-30 at 5 32 53 pm

## Removes unused function

It's a function we were never using, and it was generating a warning due to it being `public`, so I removed it. 

## Fixes storyboard warnings

Lots of warnings like this:

screen shot 2015-03-30 at 5 35 07 pm

## Fixes intermittently failing tests

This was caused by SDWebImage caching images after the first tests were run. So the first tests passed because SDWebImage was actually trying to fetch images, but it wouldn't return them in time to set the images. It would then cache them to disk, so the next time you ran the tests, SDWebImage was able to use the on-disk cache.

So we use DI to avoid SDWebImage in tests. 

## Fixes WiFi name for testing at Artsy HQ

New wifi SSID means an update to our code. 

## Removes uses of println

Fixes #397.


ashfurrow:
Currently broken; gonna amend the last commit to not be broken tomorrow.

orta:
Cool

ashfurrow:
@orta this is ready for review :space_invader:

Kiosk/Auction Listings/ListingsViewController.swift
```diff
@@ -169,7 +171,7 @@ public class ListingsViewController: UIViewController {
             developmentEnvironment = true
         #else
             if let ssid = getSSID() {
-                let developmentSSIDs = ["Ash's Wi-Fi Network", "Art.sy", "Artsy2"] as NSArray
+                let developmentSSIDs = ["Artsy"] as NSArray
```
orta: I think Artsy 2 is still useful for 27

Kiosk/Storyboards/StoryboardIdentifiers.swift
```diff
@@ -20,6 +20,11 @@ public enum SegueIdentifier : String {
     case ShowAdminOptions = "Show Admin Options"
     case ShowLoadingView = "Show Loading View"
     case ShowSaleArtworkDetails = "Show Sale Artwork Details"
+    case UNUSEDManualcreditcard = "UNUSED Manual credit card"
+    case UNUSEDPushtoSwipeCreditCard = "UNUSED Push to Swipe Credit Card"
+    case UNUSEDPushtoemail = "UNUSED Push to email"
+    case UNUSEDPushtoentermobilenumber = "UNUSED Push to enter mobile number"
+    case UNUSEDPushtopassword = "UNUSED Push to password"
```
ashfurrow: Rather than give up the visual cues helpful in the storyboard, I prefixed the identifiers with `UNUSED` to indicate that they aren't used.

Kiosk/Storyboards/StoryboardIdentifiers.swift
```diff
@@ -20,6 +20,11 @@ public enum SegueIdentifier : String {
     case ShowAdminOptions = "Show Admin Options"
     case ShowLoadingView = "Show Loading View"
     case ShowSaleArtworkDetails = "Show Sale Artwork Details"
+    case UNUSEDManualcreditcard = "UNUSED Manual credit card"
```
orta: heh

ashfurrow:
screen shot 2015-03-31 at 3 49 44 pm

Kiosk/Auction Listings/ListingsViewController.swift
ashfurrow: Oh, I didn't realize that SSID was still in use.

ashfurrow:
I'm getting to be something of a gunslinger when it comes to interactive rebases.

orta:
OK, looks good on a 2nd run through.

Issue #333, opened 2014-12-03
MVVMification


ashfurrow:
We've got a lot of heavy view controllers (looking at you, ListingsViewController). We should move the code into view models where possible, writing tests as we go. (Since our VCs are woefully undertested right now, this is a great excuse to write some!)

This should cover work from https://github.com/artsy/eidolon/issues/176 too.

PR #414, opened 2015-04-03
@orta => Removes RACSubject use in paging code

### LINKED ISSUE #331: Avoid subject in paging code
We currently do [this](https://github.com/artsy/eidolon/blob/f3c9f884035219f8590d573d06b7cdb7d16b5059/Kiosk/Auction%20Listings/ListingsViewController.swift#L49-L69) to handle paged network requests (the idea being that we keep asking for pages of n length until the length of page returned from the API is less than n). This recursive function is bootstrapped by [this one](https://github.com/artsy/eidolon/blob/f3c9f884035219f8590d573d06b7cdb7d16b5059/Kiosk/Auction%20Listings/ListingsViewController.swift#L72-L86), which creates the subject that gets passed in. 

I feel like there is almost certainly a better way to do this, but we didn't have time in the initial launch to do anything better. Might reach out to @jspahrsummers for advice.

ashfurrow:
This is really difficult to test, but this view controller is my main target for https://github.com/artsy/eidolon/issues/333 , and I promise to test it then :heart: 

Fixes #331.

### DIFF
--- a/Kiosk/Auction Listings/ListingsViewController.swift
+++ b/Kiosk/Auction Listings/ListingsViewController.swift
@@ -60,36 +60,35 @@ public class ListingsViewController: UIViewController {
     class public func instantiateFromStoryboard(storyboard: UIStoryboard) -> ListingsViewController {
         return storyboard.viewControllerWithID(.AuctionListings) as ListingsViewController
     }
-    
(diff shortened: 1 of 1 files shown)


ashfurrow:
@orta :green_apple:

orta:
:+1:

PR #432, opened 2015-04-16
@orta => Fixes form validation problems

### LINKED ISSUE #431: Users can register with invalid password
Think this is what happens:

Register with a password that is too short. Hit "Next" on the keyboard. The user is registered (despite a possible server error?) but can't log in since their password isn't long enough.

ashfurrow:
Few things:

- Login no longer requires a minimum password length.
- Consolidated use of `returnKeySignal()` to execute commands on existing buttons. These commands will not execute if they are disabled, so we avoid duplicating validation for return key presses and confirm button presses.

@orta Wasn't quite sure how to test this, so I'm abstracting the commands and enabled states into a generic view model, vis-a-vis #333.

Fixes #431.


ashfurrow:
I've added the view models. One is generic and one is specific. I'm going to add tests for them, as well as update the tests for the view controllers (now that we can inject view models, we can easily mock state and stub calls).

https://github.com/artsy/eidolon/commit/b1ee2576aadc19870e50f0861a26008211feb42a is just something I noticed – I want to make sure that if the pinging fails – we're offline, say – then we restart it.

ashfurrow:
Added tests for the new view models; going to add some basic snapshot tests for the view controllers, too.

KioskTests/Bid Fulfillment/RegistrationPasswordViewModelTests.swift
```diff
@@ -0,0 +1,287 @@
...
+import Swift_RAC_Macros
+import Kiosk
+import Moya
+
+class RegistrationPasswordViewModelTests: QuickSpec {
+    override func spec() {
+        let testPassword = "password"
+        let testEmail = "test@example.com"
+
+        // Just so providers form individual tets don't bleed into one another
```
orta: > tests

ashfurrow:
@orta Should be good to review.

ashfurrow:
:green_apple: AW YEAH

orta:
Looks good :👍

PR #434, opened 2015-04-19
[Gemfile] Update CocoaPods to 0.37.0.beta.1

segiddins:
\c @ashfurrow


ashfurrow:
CI fail is because PRs from forks don't have access to the secure variables.

PR #442, opened 2015-04-30
@orta => Fixes crash with re-binding signals

### LINKED ISSUE #441: Crashing on Edits to Registration Info
Register -> Enter some info until you get to CC -> Go back to mobile -> Hit Next -> Crash

RegistrationSubController's are being retained longer than we expect, which is causing problems binding signals from those VCs to `bidDetails.newUser` properties. The binding lasts as long as the view controller, and if it's re-binded, then it crashes. 

Solution is to have a `.takeUntil()` for some signal where it's hidden/cancelled.

ashfurrow:
If you have an idea for a better abstraction for these tests, I'd love to implement it. 

Fixes #441.

### DIFF
--- a/KioskTests/Bid Fulfillment/RegistrationMobileViewControllerTests.swift
+++ b/KioskTests/Bid Fulfillment/RegistrationMobileViewControllerTests.swift
@@ -19,5 +19,20 @@ class RegistrationMobileViewControllerTests: QuickSpec {
             subject.bidDetails.newUser.phoneNumber = "1234567890"
             expect(subject).to( haveValidSnapshot() )
         }
+
+        it("unbinds bidDetails on viewWillDisappear:") {
+            let runLifecycleOfViewController = { (bidDetails: BidDetails) -> RegistrationMobileViewController in
+                let subject = RegistrationMobileViewController.instantiateFromStoryboard(fulfillmentStoryboard)
+                subject.bidDetails = bidDetails
+                subject.loadViewProgrammatically()
+                subject.viewWillDisappear(false)
+                return subject
+            }
+
+            let bidDetails = testBidDetails()
+            runLifecycleOfViewController(bidDetails)
+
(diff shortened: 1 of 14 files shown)


KioskTests/Bid Fulfillment/RegistrationMobileViewControllerTests.swift
```diff
@@ -19,5 +19,20 @@ class RegistrationMobileViewControllerTests: QuickSpec {
             subject.bidDetails.newUser.phoneNumber = "1234567890"
             expect(subject).to( haveValidSnapshot() )
         }
+
+        it("unbinds bidDetails on viewWillDisappear:") {
+            let runLifecycleOfViewController = { (bidDetails: BidDetails) -> RegistrationMobileViewController in
+                let subject = RegistrationMobileViewController.instantiateFromStoryboard(fulfillmentStoryboard)
+                subject.bidDetails = bidDetails
+                subject.loadViewProgrammatically()
+                subject.viewWillDisappear(false)
```
orta: unless you really have to call this directly, it should probably be the `[subject beginTransition:NO animated:NO]` and `[subject endTransition]`

orta:
It's a shame that you can't check if you're already registered for KVO bindings

orta:
in terms of ideas, not really running through the `beginAppearanceTransition` functions twice is probably the closest you can get WRT making sure the view controller stuff is replicated correctly for showing then re-showing the hierarchy. Another option could be to put it in a UINavigationController, then push & pop.

KioskTests/Bid Fulfillment/RegistrationMobileViewControllerTests.swift
ashfurrow: It doesn't really matter to me – the test is that the subscriptions end when `viewWillDisappear:` is called, so that's what I called. Happy to change.

orta:
I don't feel particularly strongly about the `beginAppearanceStuff` so merging :+1:

PR #445, opened 2015-05-05
@orta => Fixes crash

### LINKED ISSUE #444: Crash with existing email address
Try to bid -> Enter a(ny) phone number -> enter existing user's email address -> Crash

ashfurrow:
Basically what we do now is end the subscription (the binding) whenever the command sends `true` on `executing`, indicating that it has started executing (we ignore `false`). Since the first invocation of the command necessarily moves to one of two new controllers, we should be good :shipit: 

Fixes #444.


orta:
:+1:

orta:
mergable on :green_apple:

ashfurrow:
Travis is being Travis-y – pushed an amended commit.

PR #447, opened 2015-05-07
Removed `team_id` as `team_name` is provided already

KrauseFx:
Please test it first, but if the name is provided correctly, there is no need to provide the ID.


ashfurrow:
Thanks!

PR #461, opened 2015-06-05
@orta => Updates to Swift 1.2/Xcode 6.3

### LINKED ISSUE #451: Update to Swift 1.2 / Xcode 6.3
Or whatever is current at the time this issue is addressed. 

This is actually gonna be a pretty big thing to do; we need to make sure that our dependencies are 1.2-ready, too. Not a big deal, except ReactiveCocoa. RAC's structure has changed since the version we're using now (new dependencies, which don't have pods yet). Given #448, it may make sense to decide whether or not we're staying with ReactiveCocoa first before/while upgrading.

ashfurrow:
Lots of things done, still a few to do. 

- [x] Updates all the Swift pods to versions compatible with Swift 1.2.
- [x] Updates eidolon's use of Moya to the new 1.0 API. 
- [x] Updates eidolon app/test code to Swift 1.2.
- [ ] Move to a CI ([Travis](http://blog.travis-ci.com/2015-05-26-xcode-63-beta-general-availability/) or Circle) that supports Swift 1.2.
- [x] Fix compiler segfaults caused by `ListingsViewControllerTests.swift` ([clue](https://github.com/Quick/Quick/blob/master/Documentation/SharedExamples.md)?).
- [x] Make tests pass. 
- [x] Smoke test the app.

Fixes #451.


ashfurrow:
So the simulator wasn't launching on CircleCI, which is a problem we've seen before on Travis. Luckily they have documentation(!) around this and recommend switching to xctool to fix it. 

So I did. Initial build on xctool failed on CI because checking the pods manifest file took too long. I've added a reporter in the hopes that maybe we just hit some kind of max log limit or something. We'll see.

ashfurrow:
Looks like it's still hanging :\

> Run custom shell script 'Check Pods Manifest.lock'

ashfurrow:
This is passing locally. 

screen shot 2015-06-16 at 5 28 27 pm

I've reached out to CircieCI about this. Our builds seem to stall at random points. I'm suspicious that we may be hitting some kind of maximum file size of the log or something. Will try a few more things, then gonna ignore CI and we can merge anyway :disappointed:

orta:
not mergable

Kiosk/App/Views/Spinner.swift
```diff
@@ -22,7 +22,7 @@ public class Spinner: UIView {
         spinner.center = CGPointMake( CGRectGetWidth(bounds) / 2, CGRectGetHeight(bounds) / 2)
     }
 
-    func animate(times: Float) {
```
orta: nice touch

orta:
as-bangs everywhere
Giphy

orta:
Looks cool to me though

ashfurrow:
Re-added Travis. The devil you know, etc. But if their Xcode 6.3 actually works, then who am I to argue?

ashfurrow:
Huh, so using xctool on Travis _also_ hangs at the [same spot](https://travis-ci.org/artsy/eidolon/builds/67241682#L5470) as CircleCI. This never happened before using xctool. Reverting to xcodebuild to see what happens.

ashfurrow:
I'm tired of mucking around with CI. Test pass locally. Orta's happy. Merging.

PR #464, opened 2015-06-17
Use StoryboardNames enum to identify storyboards instead of 'hardcoded strings'

patoroco:
Thanks to eidolon project I discovered ```sbconstants``` command, and it looks really interesting to be aware of changes in storyboards identifiers.
Using that identifiers instead of "raw strings", do that when they change, compiler tell us that something was wrong, so it's better to use these identifiers.


ashfurrow:
Cool, great idea! CI is failing just because of a pr from a fork. Thanks a lot!

orta:
nice

PR #466, opened 2015-06-18
Trying CircleCI debugging thing.

ashfurrow:
Apparently [this](https://devforums.apple.com/message/1125867#1125867) fixes the sim launch problems.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,7 +46,7 @@ GEM
       fuzzy_match (~> 2.0.4)
       nap (~> 0.8.0)
     cocoapods-downloader (0.9.0)
-    cocoapods-keys (1.2.0)
+    cocoapods-keys (1.4.0)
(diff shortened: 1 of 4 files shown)


ashfurrow:
``` no rule to process file '/Users/distiller/eidolon/Pods/CocoaPodsKeys/Keys.podspec.json' of type text.json for architecture x86_642015-07-01 09:00:43.480 
```

Update CocoaPods-Keys should fix that. The Circle people seemed to think it may be causing the problem.

orta:
That shouldn't be failing the build though, that's a warning, think it's still in there

ashfurrow:
It's a simulator launch problem – the doesn't fail, it just times out :( Bit of a hail marry :shrug:

ashfurrow:
Contacted them again – hoping to hear back soon.

PR #475, opened 2015-06-29
@alloy => Manual CC Entry Problem 📱

### LINKED ISSUE #455: Swallowed error on entering cc
Witnessed a user register, enter a cc, then enter an expiration date, then hit ok and get nothing back. Spinner spinned, then the button came back clickable without any errors. CC seems to have been entered into the system. The only workaround was to take someone's bid on an ipad. Trying to bid again on a kiosk was asking for the cc again, didn't retry.

ashfurrow:
Crus is that we weren't sending a "complete" event to our `finishedSignal` once the CC tokenization succeeded. I added tests to ensure that the event is sent if and only if the tokenization succeeds. 

I also noticed a few other problems that I fixed and added test for, in separate commits. Should I make things like that separate PRs? Open to advice.

(Note: Circle and Travis are currently both failing to run tests. I've checked that they pass locally.)

Fixes #455.

### DIFF
--- a/Kiosk.xcodeproj/project.pbxproj
+++ b/Kiosk.xcodeproj/project.pbxproj
@@ -721,8 +721,6 @@
 				5E82BD341AF2C6E700CB02DD /* YourBiddingDetailsViewControllerTests.swift */,
 				5E8610CF1A5C43EF00456E41 /* PlaceBidViewControllerTests.swift */,
 				5EDD04D11AA69D2700B30AE9 /* KeypadViewModelTests.swift */,
-				5EF71E821AE9B6A80069A266 /* StripeManagerTests.swift */,
-				5EF71E861AE9B6EC0069A266 /* ManualCreditCardInputViewModelTests.swift */,
 			);
 			path = "Bid Fulfillment";
 			sourceTree = "<group>";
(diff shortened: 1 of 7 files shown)


alloy:
Tested for real/device?

ashfurrow:
I've not tested it on a device yet; will do after lunch.

ashfurrow:
(Thanks for the reminder!)

alloy:
Alrighty then, I’ll wait to hear when you tested it on device.

Regarding splitting off PRs, I’m ok with this personally. If you feel wrong and it would keep you up at night you could just do the fix and then a new PR with all the ‘UI’ issues you came across.

ashfurrow:
Separate PRs are hard since sometimes the fixes I do depend on the other code I've already changed for the main issue. Gonna keep it like this, in small commits :+1:

alloy:
@ashfurrow To be clear, I meant that I’m personally ok with not splitting off, i.e. as-is.

ashfurrow:
@alloy verified on device.

alloy:
:100: :+1:

alloy:
Oh, forgot to notice before, but CHANGELOG?

ashfurrow:
Knew there was something I was forgetting.

alloy:
:100: +1 :+1:

PR #476, opened 2015-06-29
@alloy => Adds long tap gesture for bidding error details 📱

### LINKED ISSUE #454: When you bid on a work that has just sold you get Bid Failed
img_0003

A work that is already sold via buy now just displays "Bid Failed" when you try to bid on it. Really it should say the artwork has been already sold.

Generally we need an explanation next to Bid Failed which is useful. Maybe I can tap and hold the message and get details instead of a generic message because when this happens we have an irate user standing there and a powerless Artsy person.

ashfurrow:
Adds a tap gesture recognizer to the error details label that will display an alert with the technical details, including the API response HTTP status code (if any), body length, and UTF8-encoded body string (if any).

I've produced an intentional error to generate the screenshot below:

ios simulator screen shot jun 29 2015 3 23 18 pm

Tests pass locally.

Fixes #454.


alloy:
Nice! :+1:

ashfurrow:
:bow:

alessandra-castillo:
Thanks Ash!!

PR #477, opened 2015-07-01
@orta => Disables bid buttons for sold artworks 📱

### LINKED ISSUE #453: When a work is sold (via buy now) at an auction it needs to say so
If a work at auction is bought via "buy now", it should say Sold in the UI. Right now you can still bid on it.

ashfurrow:
Disables the button based on the `sold` field of the artwork JSON. As discussed in #453, the heuristic is "sold" means sold and anything else means not sold. Seems reasonable to me. 

Note: this puts Eidolon over the 200 count for unit tests :tada: A lot of those are shared examples, which have really grown on me, since they enable a way to test code under breadth of different conditions. 

Fixes #453.


Kiosk/App/Models/Artwork.swift
```diff
@@ -30,13 +46,14 @@ public class Artwork: JSONAble {
         return defaultImages?.first ?? self.images?.first
     }()
 
-    init(id: String, dateString: String, title: String, titleAndDate: NSAttributedString, price: String, date: String) {
+    init(id: String, dateString: String, title: String, titleAndDate: NSAttributedString, price: String, date: String, sold: String) {
```
orta: just to play dumb here, but I've been doing a bit of that there swift lately, and it would automatically create these full init functions for me ( I was using structs though ) - can it work for this?

KioskTests/ListingsViewControllerTests.swift
```diff
@@ -73,6 +73,16 @@ class ListingsViewControllerTests: QuickSpec {
                 }
                 itBehavesLike("a listings controller") { ["subject": subject] }
             }
+
+            describe("with artworks not for sale") {
+                beforeEach {
+                    subject.beginAppearanceTransition(true, animated: false)
+                    subject.endAppearanceTransition()
+                    subject.saleArtworks.map { $0.artwork.soldStatus = "sold" }
```
orta: this is nice, I agree

orta:
A lovely PR. 

Travis just updated to Xcode 8.4 BTW, https://twitter.com/travisci/status/616007115832999936

Kiosk/App/Models/Artwork.swift
ashfurrow: That's a good question. Classes do allow you to do this, but only for `var` properties that have default values (it'll generate an `init()` function). So this wouldn't work:

```swift
class Thing {
    let t: String
}

let thing = Thing(t: "hi") // 😓
```

However, _structs_ will do this fine. 

```swift
struct Thing {
    let t: String
}

let thing = Thing(t: "hi")
```

ashfurrow:
Thanks! :bow:

Both Travis and Circle are having problems launching the simulator to actually run the tests. Update to Xcode 6.4 may help. I've been [working with Circle](https://github.com/artsy/eidolon/pull/466) to troubleshoot the problem, but it's on the back burner.

PR #478, opened 2015-07-01
@orta => Adds Questions section to readme

ashfurrow:
Need to let people know we're approachable.


orta:
"or via mobile@artsymail.com" or whatever we put on the eigen release notes?

ashfurrow:
Good idea

ashfurrow:
Actually, I like the idea of keeping things in the repo. If someone sends an email, then our reply is just to them instead of the world. These are almost certainly developers, not users, so they'll likely have technical questions or feedback.

PR #480, opened 2015-07-02
@orta => Updates existing artworks sold status

ashfurrow:
I realize that I made an error on [this pull request](https://github.com/artsy/eidolon/pull/477). Artworks that were sold when the app first sync'd would be marked as sold, but artworks that were updated after the first sync would not. 

I don't think it needs a changelog entry since we've not released since the PR was merged. My bad for not thinking of this scenario at the time.


orta:
Looks legit to me. Presuming CI fails.

PR #485, opened 2015-07-24
@orta: Adds new backend estimateCents field to SaleArtwork model

### LINKED ISSUE #474: Kiosk displaying "No Estimate" when estimate field is actually filled
New "Estimate" field was recently added to the Auctions App and first used in the amfAR generationCURE auction.
screen shot 2015-06-25 at 4 31 00 pm

My guess is that the kiosk doesn't know to pull estimates from that new field, and instead displays "No Estimate." 
img_0004 2

ashfurrow:
I went looking on Google for "art auction" GIFs and was happy to see we're on the first page:

screen shot 2015-07-24 at 2 57 19 pm

/cc @dblock 

Fixes #474.

### DIFF
--- a/Kiosk/App/Models/SaleArtwork.swift
+++ b/Kiosk/App/Models/SaleArtwork.swift
@@ -40,6 +40,7 @@ public class SaleArtwork: JSONAble {
     public dynamic var minimumNextBidCents: NSNumber?
     
     public dynamic var highestBidCents: NSNumber?
+    public var estimateCents: Int?
     public var lowEstimateCents: Int?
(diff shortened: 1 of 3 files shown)


Kiosk/App/Models/SaleArtwork.swift
```diff
@@ -93,17 +96,18 @@ public class SaleArtwork: JSONAble {
     }
     
     public var estimateString: String {
+        // Default to estimateCents
```
orta: I read this as Esticents

orta:
cool, looks good.

Issue #396, opened 2015-03-11
Update List view


katarinabatina:
Changes that match web visual styles

![image](https://cloud.githubusercontent.com/assets/6482926/6606684/cfdd360e-c80f-11e4-951f-ed3ab2a0f3f7.png)

ashfurrow:
We'll also want to verify that we order by lot number – probably are already, but the existing design doesn't guarantee that.

PR #486, opened 2015-07-24
@orta: Fixes crash on grid view

### LINKED ISSUE #473: Kiosk crashes after selecting "Grid View" from any other lot view
:(

ashfurrow:
If an aspect ratio is not specified, square is used instead. Seems appropriate since we're updating them all to be square in #396 anyway. But to be safe, I removed one of the aspect ratios in our test data. 

I noticed this was happening on launch for the Watermill auction, too. Confirmed it's working with amFar data, too. 

Confirmed tests are passing:

screen shot 2015-07-24 at 4 24 31 pm

Fixes #473.


orta:
Cool!

PR #487, opened 2015-07-24
@orta: Tapping anywhere in the list cell pushes the detail VC

### LINKED ISSUE #391: Make the entire row clickable in grid views
I witnessed a user at LAMA desperately tap on the line for a lot in the grid view (the A-Z view I think) until (after no less than 10 tries and almost giving up) they finally saw the bid button and clicked on that. I think Bid should lead you to an actual bid page while tapping on the work or anywhere else on the line should lead you to more focused info.

ashfurrow:
Fixes #391.


orta:
:+1:

PR #488, opened 2015-07-24
@orta: Displays thumbnail on detail view while fullsize image is being downloaded

### LINKED ISSUE #429: Detail image views
We show a grey background until the full sized image is loaded. We should show the thumbnail instead.

ashfurrow:
Fixes #429.


orta:
:+1: - might also make sense to do this as a category, I think we do that in folio/eigen

PR #496, opened 2015-08-25
@orta: Swift 2

ashfurrow:
Initial work, ongoing. 

Ran the converter, which failed shortly after beginning. 

Updated our dependencies – more to come.

### DIFF
--- a/Kiosk/Auction Listings/ListingsViewController.swift
+++ b/Kiosk/Auction Listings/ListingsViewController.swift
@@ -94,7 +94,7 @@ public class ListingsViewController: UIViewController {
             
             let array = object as? Array<Array<AnyObject>>
             return reduce(array ?? [], Array<AnyObject>(), +)
-        }).mapToObjectArray(SaleArtwork.self).catch({ (error) -> RACSignal! in
+    }).mapToObjectArray(SaleArtwork.self).`catch`({ (error) -> RACSignal! in
             
             logger.log("Sale Artworks: Error handling thing: \(error.artsyServerError())")
 
--- a/Kiosk/Auction Listings/ListingsViewController.swift
+++ b/Kiosk/Auction Listings/ListingsViewController.swift
@@ -74,7 +74,7 @@ public class ListingsViewController: UIViewController {
 
                     var nextPageSignal = RACSignal.empty()
 
-                    if count(array) >= (self?.pageSize ?? 0) {
+                    if array.count >= (self?.pageSize ?? 0) {
                         // Infer we have more results to retrieve
                         nextPageSignal = self?.retrieveAllListingsRequestSignal(auctionID, page: page+1) ?? RACSignal.empty()
                     }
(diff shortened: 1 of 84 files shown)


ashfurrow:
screen shot 2015-08-25 at 1 41 56 pm

:cry:

ashfurrow:
The remaining app code is updated so that it _compiles_, but of course the app isn't behaving properly :stuck_out_tongue: Also, I still need to get the unit tests going.

ashfurrow:
OK, so unit tests are _compiling_, but we have **49** out of 224 tests failing. Most of those are snapshot tests, and most of _those_ are off-by-one-pixel problems, and are easily fixed. Hopefully this will be ready for review this afternoon.

ashfurrow:
Unit tests pass and app runs as expected on iOS 8.1, but fail on iOS 9. 

Possible explanations:

- iOS 9 has subtly changed view lifecycles, causing the empty button failures we noticed.
- iOS 9 is still in beta and can't be relied upon.

Kiosk/Auction Listings/ListingsViewController.swift
```diff
@@ -114,18 +114,18 @@ public class ListingsViewController: UIViewController {
...
-                let sortedNewSaleArtworks = newSaleArtworks.sorted(sortById)
+                let sortedCurentSaleArtworks = currentSaleArtworks.sort(sortById)
+                let sortedNewSaleArtworks = newSaleArtworks.sort(sortById)
                 
-                let saleArtworksCount = count(sortedCurentSaleArtworks)
+                let saleArtworksCount = sortedCurentSaleArtworks.count
                 for var i = 0; i < saleArtworksCount; i++ {
                     if currentSaleArtworks[i].id == newSaleArtworks[i].id {
-                        currentSaleArtworks[i].updateWithValues(newSaleArtworks[i])
+                        currentSaleArtworks[i].updateWithValues(sortedNewSaleArtworks[i])
```
ashfurrow: This was always broken – Swift compiler warning caught it.

KioskTests/Bid Fulfillment/ConfirmYourBidPINViewControllerTests.swift
```diff
@@ -35,14 +34,10 @@ class ConfirmYourBidPINViewControllerTests: QuickSpec {
 
             customKeySubject.sendNext("2");
             expect(subject.pinTextField.text) == "2"
-
-            subject.keypadContainer.deleteCommand.execute(nil)
-            expect(subject.pinTextField.text) == ""
```
ashfurrow: I don't know how this (or the test that follows) _ever_ passed. It relies on listening to the `pinSignal` for updates, which we want because it should send "" when we invoke `deleteCommand`. **However**, we explicitly replace the subject's `pinSignal`, so it'll _never_ send "" :confused:

ashfurrow:
OK, I've looked this over and I'm satisfied more or less. Going to review with @orta tomorrow.

orta:
We've reviewed in person on the same computer. :+1: 

MRW I know something exciting is going on with Black Holes, but not really sure WHAT, so I go in and start reading the Event Horizon Telescope Project Astrophysicists AMA.

PR #513, opened 2015-09-28
@orta: Lot number fix

ashfurrow:
This fixes a bug tracked in [this Trello card](https://trello.com/c/f3GvqhJl/82-kiosk-sync-using-wrong-lot-numbers). Here's what was going on:

- We had an auction that changed the original lot numbers. 
- The initial sync displayed the correct lot details. 
- Subsequent syncs would show incorrect details. 

Here's what I think was happening (@devangt is going to confirm this afternoon):

1. On first sync, the lengths of the new sale artworks and current sale artworks differ (the latter is empty), so we just use the new data. 
2. On subsequent syncs, we checked if the arrays lengths differed (let's assume they don't). We were then sorting the new/current sale artworks by `id`, then marching through the two arrays (of equal length). If the `id`s at each index matched, then we updated the existing `SaleArtwork` instance in situ. 

The problem with step 2 is that we were updating the existing `SaleArtwork` at the index of the _unsorted_ array with the sale artwork at the same index in the _sorted_ array. Since the default sort returned by the API is by lot number, when we swapped the lot numbers of the first and second lots, subsequent syncs would apply the updates for the first lot to the second one, and vice-versa. 

The fix involves _not sorting_ at all – relying on the default sort order of the API (lot number), and if any of the IDs of the sale artworks have changed, then we replace the entire array.

Wanted to get this PR up to run this by others. **Before merging**, this needs:

- [x] Tests for breaking case.
- [x] Changelog update.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -1,4 +1,7 @@
-upcoming: 
+upcoming:
+  version: 3.6.1
+  notes:
+  - Fixes problem with lot numbers changing during the auction [ash]
 releases:
   date: September 24 2015
   version: 3.6.0
--- a/Kiosk/App/Models/SaleArtworkViewModel.swift
+++ b/Kiosk/App/Models/SaleArtworkViewModel.swift
@@ -52,6 +52,10 @@ extension SaleArtworkViewModel {
         return saleArtwork.artwork.titleAndDate
     }
 
+    var saleArtworkID: String {
+        return saleArtwork.id
+    }
+
     // Signals representing values that change over time.
 
     var numberOfBidsSignal: RACSignal {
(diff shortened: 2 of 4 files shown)


orta:
Cool, this looks good. Serious test that one.

ashfurrow:
Yeah, took a while to get the tests right. Might be the jet lag, but working through the tests led me to a simpler solution altogether.

ashfurrow:
:green_apple:

orta:
awesome, and are we still :+1:  on the same size assumption? if so merge away.

ashfurrow:
Yeah, there's a check for it beforehand. Should probably move it to `guard` :smile:

PR #520, opened 2015-10-05
@ashfurrow: Auctions 106 poll bidder position

acjay:


acjay:
Ah, better. I may need some help testing on my simulator. I don't think I'm set up with the right API keys.

acjay:
Just kidding, didn't pass

ashfurrow:
Yeah, so the setup for Artsy people is: clone, bundle install, then pod install. At this point you'll be asked for the API keys, which are in the 1Password. Make sure you debase from master, with the changes to the podfile. Happy to merge this manually for now, too.

ashfurrow:
The test failures are likely related to the fonts being different, too.

Kiosk/App/Networking/ArtsyAPI.swift
```diff
@@ -303,6 +307,9 @@ extension ArtsyAPI : MoyaTarget {
 
         case MyBidPositionsForAuctionArtwork:
             return stubbedResponse("MyBidPositionsForAuctionArtwork")
+            
+        case MyBidPosition:
+            return stubbedResponse("MyBidPosition")
```
ashfurrow: There's a directory with other `.json` files that needs to have a file in it that is named `MyBidPosition.json`, which contains a stubbed network response, otherwise the tests will crash.

ashfurrow:
Cool, I corrected the podfile. I'll send a PR with updated instructions for Artsy developers. Let me know if I can help!

ashfurrow:
I've opened a PR for documentation fixes here: https://github.com/artsy/eidolon/pull/521 Your API keys are still stubbed, which is a problem. The plugin we're using to manage keys doesn't have a way to delete them all (/cc @orta) so the easiest solution is honestly to rename the directory that you've checked the code into.

acjay:
When I run `bundle exec pod install`, it keeps switching to the OSS fonts. Is there any way to curb this?

ashfurrow:
That's strange – have you pulled in from `master`? I remember we had to change the `Podfile` to reflect your username on your computer.

acjay:
Yep, that's the problem. I'm working on the other branch that doesn't have that change.

ashfurrow:
Merged that, let me know how it goes.

acjay:
Yep, fixed :+1:

ashfurrow:
Awesome :100: Looks like the test failure is legit – let me know if I can assist.

acjay:
Alright, I've got staging working, although I can't really test this without placing a bid, but can't place a bid without registering, but can't register without #520. We already established during today's mocktion that registering elsewhere with a passing ZIP/CVV still won't let us authenticate by paddle+pin, for some reason.

acjay:
Actually, it did just work this time to place a bid after registering through Force. But there's the faulty Place A Higher Bid button. I was hoping that was a problem that would go away with this fix.

screen shot 2015-10-07 at 6 26 35 pm

acjay:
Looking at Ohm, I'm not the highest bidder, so the Place Higher Bid button is correct, but I would think the message would be more accurate, especially since I'm pretty sure the polling didn't expire. My logic could be faulty.

ashfurrow:
Yeah, that message on the Kiosk indicates that your bit was submitted, but polling failed to indicate if you were the highest bidder.

ashfurrow:
Cool, so https://github.com/artsy/eidolon/commit/2c35f1a80780d2a32582690d4f02f7426e73b6a5 includes a fix for the text. The stub view model was no longer overriding a method that the tests were expecting to be called. Gonna wait for :green_apple: to merge, then I'll do a deploy.

acjay:
(fingers crossed)

ashfurrow:
You've tested on staging that this works as expected, eh? I can if you like (I just wanna the unit tests to pass :stuck_out_tongue:)

acjay:
Haven't, due to not knowing until today that we have the admin buttons for making the swipe work. I was thinking we'd have to merge the CVV one first to be testable. I think we can't really confirm that it fixes the original problems, since they're transient, so it'll be sufficient if we can make sure regular bidding works as expected and is generally giving us expected messages.

acjay:
Huzzah! Thanks!

ashfurrow:
Gotcha – ok, going to do a deploy and you'll let me know from the mock auction tomorrow how it goes?

acjay:
Indeed. Although we'll want to get that other one in too. Should be a matter of updating the snapshots i think. Probably should add more tests as well, matching the existing ones, but I think that can wait.

acjay:
Do you have bandwidth to update the failing snapshot test? Meantime, I'll give this a spin in the simulator to make sure we're looking as expected.

ashfurrow:
Other one is already in – updated the snapshot tests. Agreed we need more tests, but right now shipping is more important.

acjay:
You have saved all our lives, thanks sir!

PR #522, opened 2015-10-07
@ashfurrow : add CVV to manual card entry

acjay:
Due to my simulator seemingly not connecting to staging auctions, I haven't been able to actually try this on my own, so it may not quite work. I tried to write it completely analogously to existing code.


ashfurrow:
Cool, about to hop on a plane – will take a look when I land. How did the other PR go, for polling?

acjay:
Fixed one problem here but there's still another problem of a couple bad snapshots. Not sure how we update those.

ashfurrow:
Easiest way to update them is to use @orta's Xcode plugin: http://github.com/orta/snapshots/

acjay:
Whoa, plugins! This could be a gamechanger. What do you and @orta recommend?

ashfurrow:
Nothing besides the one that Orta made – he probably has other recommendations :)

ashfurrow:
Should be able to pull the same trick with rebasing as #520 to test against staging.

acjay:
Alright fixed the keys and the Podfile, but still can't see the live test auction

ashfurrow:
That is so bizarre! Just to confirm, you did a rebase from master, then a `pod install`, and then ⌘R to re-run the app?

ashfurrow:
That is so bizarre! Just to confirm, you did a rebase from master, then a `pod install`, and then ⌘R to re-run the app?

acjay:
I merged master, but yes

ashfurrow:
And a `pod install` prompted you for keys?

acjay:
Yep, I entered all the most recent keys

ashfurrow:
(For the record, Alan and I resolved this in Slack – a clean build in Xcode solved things.)

ashfurrow:
I've fixed most of these – just two failing snapshots I'm looking at now.

ashfurrow:
OK so this should be good to go. @acjay feel free to revert this PR if the changes ( https://github.com/artsy/eidolon/commit/14db2b64891278731f4b6bd9c7d55bec99ded7a5 ) don't make sense – I only had to modify a few tests, and remove one that no longer made sense with CVV entry. 

During the next mock auction, I'd appreciate it if someone could swipe an actual credit card on production (just registering!) to verify 100% that the CVV stuff works against Stripe production and the APIs changes for trusting CCs.

PR #528, opened 2015-10-12
@ashfurrow => raise bid keypad limit to 10,000,000

acjay:


acjay:
In an ideal world, we'd perhaps not hardcode any limit here, but have the limit be determined by the central rules, whatever they may be, and the interface capable of adapting. Of course, there are practical limits to what numbers fit within the design, if bids are denominated in, say, Satoshis :)

ashfurrow:
Ha, yeah. @joeyAghion and I have discussed this, I think. It's basically just so that we have _some_ limit that we're aware of, instead of accidentally hitting some sort of integer overflow during an event or something.

PR #530, opened 2015-10-13
[WIP] @acjay: Fixes PIN login failures

ashfurrow:
This fixes log in with bidder/phone number and pin. The issue was that the XApp token requested and stored by the original provider wasn't being added for providers created for logged-in users. This didn't affect users logged in with Artsy credentials because they had their own auth done with username/password which is separate from the trust token endpoint. 

This still needs tests, so it's WIP for now.

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -2,6 +2,7 @@ upcoming:
   notes:
   - Increases max bid to 10 000 000 [alan]
   - New safeguard to ensure Kiosk is hitting network [ash]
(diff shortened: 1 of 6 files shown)


acjay:
Looks good to me! Besides the logging code, this seems to subtract characters, which always feels like a good sign :) I don't currently fully understand Moya or our auth scheme, so I may not be the best person to judge. Maybe you can walk me through when we meet to talk deployment process.

orta:
Auth looks good from my perspective, at least WRT the changes + explanation :+1:

PR #532, opened 2015-10-13
@acjay: "Place Higher Bid" button shown in case of sniping.

ashfurrow:


ashfurrow:
So frustrating – pretty sure these are just intermittent failures, since they're succeeded locally.

PR #534, opened 2015-10-15
@orta: Reader Updates

ashfurrow:
Removes some dead code, removes reference cycles, and terminates signal when the VC disappears (cancelling or using a test CC token would make the signal persist, forever).


orta:
looks good, ships

ashfurrow:
CI will likely fail on this, but I've tested locally fine. We're tracking CI problems on Trello.

PR #535, opened 2015-10-19
@acjay: Gets CI 🍏

ashfurrow:


ashfurrow:
Oh interesting – looks like by marking the test as pending, the next one failed instead. It's like the Nth test always failed. Going to close this.

PR #539, opened 2015-10-21
@orta: Fixes problem not displaying credentials

ashfurrow:
If a new user is outbid, we allow them to increase their bid or "Continue" to view credentials. If they _did_ increase their bid, they'd never be shown those credentials because we were persisting the state of "newly registered" in a view model that never got forwarded. 

This should also address the CI issues we've been having – too many never-terminated recurring signals :sweat_smile:

### DIFF
--- a/CHANGELOG.yml
+++ b/CHANGELOG.yml
@@ -1,8 +1,12 @@
 upcoming:
+  version: 3.8.5
+  notes:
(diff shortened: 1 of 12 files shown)


orta:
aww, shame bout them reds

ashfurrow:
OK, well the tests pass _locally_ away.

orta:
Looks cool to me :+1:

PR #541, opened 2015-10-21
@orta: Better automation

ashfurrow:
This PR freshens up our use of Fastlane. 

Instead of deriving the new version from the changelog, we now require it as a parameter to `fastlane deploy`. 

```sh
fastlane deploy version:1.2.3
```

This frees us from single commits _just_ updating our changelog in preparation for a deploy. It _also_ lets us modify the changelog as part of the deploy, moving the `upcoming` notes to a new entry in the `releases` list. (I also added some spaces to make it more readable.)

I also moved the key-verification to a custom action, instead of something done before every lane. We should consider publishing the action so others can use it, too. Not sure of the work involved – will investigate. 

Finally, in the case of an error, we now remove build artefacts and reset the git repo. I've had problems with `sigh` failing due to Apple's server problems, but at that point, I've already modified my build numbers/versions, but I can't re-run the deploy without a clean `git` status. If an error occurs on _another_ lane, don't worry about losing work! `reset_git_repo` won't execute unless the git repo was checked to be clean first, so no data case be lost. 

I'm sure there are more opportunities for us to automate things, but I feel like Eidolon is becoming a good Fastlane citizen. In hindsight, using YAML wasn't _that_ big of an improvement. Moya uses markdown and has [almost the same code](https://github.com/Moya/Moya/blob/d4339ab54b5c8f7b1c43a6fa71335c2f6313e847/Rakefile#L45-L49) to update the changelog. 

/cc @KrauseFx


ashfurrow:
This CI is greeeen??? Whaaa?

ashfurrow:
Ah, I see we're back to _intermittent_ failures instead of consistent ones.

orta:
Good job we have two then, merging!

fastlane/actions/verify_keys.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Actions
+
+    class VerifyKeysAction < Action
+      def self.run(params)
+
+        # Ensure no deploys with OSS keys
+        require 'cocoapods-core'
+        puts "Validating CocoaPods Keys"
+
+        podfile = Pod::Podfile.from_file "Podfile"
```
KrauseFx: This integration looks really cool - it looks like it's useful for everyone using CocoaPods Keys? Would you want to add this to the built-in fastlane actions?

ashfurrow: Whoa, maybe! I mean, it imposes some restrictions on valid keys (they must not be a `-`) but I think that's reasonable. If I follow the instructions produced in `fastlane new_action`, that'll do it?

KrauseFx: @ashfurrow Basically just forking the fastlane repo, adding the action to `actions` folder and adding documentation to the action and to `Actions.md` does the trick :+1:

KrauseFx: Also, tests in the `specs/actions` folder :rocket:

PR #542, opened 2015-10-22
@acjay: Fixes staging flag.

ashfurrow:


acjay:
CI still busted?

ashfurrow:
Yes, tracked in Trello.

ashfurrow:
Oh weird, hold up.

Kiosk/Auction Listings/ListingsViewController.swift
```diff
@@ -44,7 +44,7 @@ class ListingsViewController: UIViewController {
 
         // Set up development environment.
 
-        if APIKeys.sharedKeys.stubResponses {
+        if APIKeys.sharedKeys.stubResponses && AppSetup.sharedState.isTesting == false {
```
acjay: As a future refactor thought, if this logic isn't strictly specific to the view, I like the idea of encapsulating it in a "config model" of some sort. That way, if anything else needs to perform a similar mode check, the actual logic is all in one spot.

ashfurrow: Agreed – this was just to get things fixed now, will follow-up.

acjay:
Holding up! Let me know when I can press the button

ashfurrow:
OK, tests passing locally. Had to change a few things. Agreed we need a better abstraction, but this will do for now. Merge when ready :+1:

ashfurrow:
Ha, nice. The deploy got rolled into the PR.

PR #545, opened 2015-10-26
@acjay: Discounts retracted single bids from sorting.

ashfurrow:
Replicates logic from here: https://github.com/artsy/eidolon/blob/49e798a272727869dabb71a0e4755891fb1c0679/Kiosk/App/Models/SaleArtworkViewModel.swift#L72-L75 I think we should move these into some sort of computed property on `SaleArtwork` or `SaleArtworkViewModel`, but this works for now. 

Fixes bug tracked in Trello: https://trello.com/c/E5faR3wq/169-kiosk-lots-that-had-bids-retracted-are-still-listed-under-highest-bids-as-if-they-still-have-a-bid

### DIFF
--- a/Kiosk/Auction Listings/ListingsViewModel.swift
+++ b/Kiosk/Auction Listings/ListingsViewModel.swift
@@ -267,7 +267,16 @@ class ListingsViewModel: NSObject, ListingsViewModelType {
 // MARK: - Sorting Functions
 
 func leastBidsSort(lhs: SaleArtwork, _ rhs: SaleArtwork) -> Bool {
-    return (lhs.bidCount ?? 0) < (rhs.bidCount ?? 0)
+    switch (lhs.highestBidCents.hasValue, rhs.highestBidCents.hasValue) {
+    case (true, true): // Both valid, compare bidCount.
+        return (lhs.bidCount ?? 0) < (rhs.bidCount ?? 0)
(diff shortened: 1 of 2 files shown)


acjay:
Uf, I think this may be continued pollution from not simplify redefining the API's concept of bid count to only reflect non-retracted bids, but until we change that, this will do.

Kiosk/Auction Listings/ListingsViewModel.swift
```diff
@@ -267,7 +267,16 @@ class ListingsViewModel: NSObject, ListingsViewModelType {
 // MARK: - Sorting Functions
 
 func leastBidsSort(lhs: SaleArtwork, _ rhs: SaleArtwork) -> Bool {
-    return (lhs.bidCount ?? 0) < (rhs.bidCount ?? 0)
+    switch (lhs.highestBidCents.hasValue, rhs.highestBidCents.hasValue) {
```
acjay: I'm a big fan of this style of logic -- `swtich` on all the data required to make a decision

ashfurrow: :bow: Getting to love it the more I use it.

ashfurrow:
Agreed – if we could possible push this up away from the clients altogether, I'm for it.

PR #546, opened 2015-10-27
@acjay: Fixes Most Bids tab problem.

ashfurrow:
The fixes in #545 missed a crucial problem: sorting by # of bids ascending and descending can no longer be done with the same sort function (negated, in one case). This was causing the bug to instead show up in the "Most Bids" tab.

screen shot 2015-10-26 at 8 59 24 pm

I know this is getting ridiculous – I'll open an issue to abstract this properly later.


ashfurrow:
Opened https://github.com/artsy/eidolon/issues/547

PR #549, opened 2015-10-27
@acjay: Reverts client-side fixes for bid counts

### LINKED ISSUE #547: Better Abstraction for bidCount
Bids don't count if they have been cancelled, which normally doesn't affect our UI in any way. However, if the _only_ bid was cancelled, we show "0 bids" instead of the slightly awkward "1 bid" because that one bid is cancelled. 

So we have special logic [here](https://github.com/artsy/eidolon/blob/49e798a272727869dabb71a0e4755891fb1c0679/Kiosk/App/Models/SaleArtworkViewModel.swift#L72-L75) and [here](https://github.com/artsy/eidolon/pull/545) and [here](https://github.com/artsy/eidolon/pull/546) to deal with this. It should be centralized to either the `SaleModel` or the `SaleViewModel`. Or pushed completely to the server side as @acjay recommended.

ashfurrow:
Fixes #547.


ashfurrow:
Tests pass locally: 

screen shot 2015-10-27 at 1 09 16 pm

Issue #433, opened 2015-04-17
Consolidate Provider usage


ashfurrow:
Throughout the app, we're using `XAppRequest()` and `Provider.sharedProvider.request()` differently. The first calls the second, after making sure we have a valid XApp token. We need to use _only_ that method through some clever abstraction.

PR #550, opened 2015-10-27
@orta: Addresses memory leaks

### LINKED ISSUE #500: Get CI working

ashfurrow:
Tried a fix for #500 but instead just got rid of reference cycles.

### DIFF
--- a/Kiosk/Auction Listings/ListingsCountdownManager.swift
+++ b/Kiosk/Auction Listings/ListingsCountdownManager.swift
@@ -23,7 +23,7 @@ class ListingsCountdownManager: NSObject {
         }
     }
 
-    /// Immediately invalidates all timers. No further updates will be made to the UI after this method is called.
+    /// Immediately invalidates the timer. No further updates will be made to the UI after this method is called.
     func invalidate() {
         _timer?.invalidate()
     }
--- a/Kiosk/Auction Listings/ListingsCountdownManager.swift
+++ b/Kiosk/Auction Listings/ListingsCountdownManager.swift
@@ -10,7 +10,9 @@ class ListingsCountdownManager: NSObject {
     dynamic var sale: Sale?
 
     let time = SystemTime()
-    
+
+    private var _timer: NSTimer? = nil
+
     override func awakeFromNib() {
         super.awakeFromNib()
         formatter.minimumIntegerDigits = 2
@@ -19,6 +21,11 @@ class ListingsCountdownManager: NSObject {
             self?.startTimer()
             self?.setLabelsHidden(false)
         }
+    }
+
+    /// Immediately invalidates all timers. No further updates will be made to the UI after this method is called.
+    func invalidate() {
+        _timer?.invalidate()
     }
 
     func setFonts() {
@@ -50,6 +57,8 @@ class ListingsCountdownManager: NSObject {
         let timer = NSTimer(timeInterval: 0.49, target: self, selector: "tick:", userInfo: nil, repeats: true)
         NSRunLoop.mainRunLoop().addTimer(timer, forMode: NSRunLoopCommonModes)
 
+        _timer = timer
+
         self.tick(timer)
     }
     
@@ -68,6 +77,7 @@ class ListingsCountdownManager: NSObject {
                 self.countdownLabel.text = "CLOSED"
                 hideDenomenatorLabels()
                 timer.invalidate()
+                _timer = nil
             }
         }
     }
--- a/Kiosk/Bid Fulfillment/BidderNetworkModel.swift
+++ b/Kiosk/Bid Fulfillment/BidderNetworkModel.swift
@@ -1,6 +1,5 @@
 import Foundation
 import ReactiveCocoa
-import Swift_RAC_Macros
 import Moya
 
 class BidderNetworkModel: NSObject {
--- a/Kiosk/Bid Fulfillment/BidderNetworkModel.swift
+++ b/Kiosk/Bid Fulfillment/BidderNetworkModel.swift
@@ -5,7 +5,7 @@ import Moya
 class BidderNetworkModel: NSObject {
     // MARK: - Getters
 
-    var fulfillmentController: FulfillmentController
+    unowned let fulfillmentController: FulfillmentController
 
     var createdNewUser: RACSignal {
         return RACObserve(self.fulfillmentController.bidDetails.newUser, "hasBeenRegistered").takeUntil(rac_willDeallocSignal())
--- a/Kiosk/Bid Fulfillment/BidderNetworkModel.swift
+++ b/Kiosk/Bid Fulfillment/BidderNetworkModel.swift
@@ -9,7 +9,7 @@ class BidderNetworkModel: NSObject {
     var fulfillmentController: FulfillmentController
 
     var createdNewUser: RACSignal {
-        return RACObserve(self.fulfillmentController.bidDetails.newUser, "hasBeenRegistered")
+        return RACObserve(self.fulfillmentController.bidDetails.newUser, "hasBeenRegistered").takeUntil(rac_willDeallocSignal())
     }
 
     init(fulfillmentController: FulfillmentController) {
@@ -56,8 +56,8 @@ class BidderNetworkModel: NSObject {
             logger.log("Creating user failed.")
             logger.log("Error: \(error.localizedDescription). \n \(error.artsyServerError())")
 
-        }.then {
-            self.updateProvider()
+        }.then { [weak self] in
+            self?.updateProvider() ?? RACSignal.empty()
         }
     }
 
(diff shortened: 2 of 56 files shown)


ashfurrow:
Becoming more a general-purpose performance PR. I'm OK with this.

Really glad when we move to a FRP lib that uses the compiler to warn us we forgot `.takeUntil(viewWillDisappearSignal())`.

ashfurrow:
So basically, there are problems with `fulfillmentNavigatonController` being retained everywhere. I've been setting a breakpoint in the nav controller's `deinit` function to make sure it's called under different scenarios. Here's the state of this PR right now:

## Broken

Bid -> Bid Button -> Paddle/Pin -> Bid -> [Cancel]
Bid -> Bid Button -> Paddle/Pin -> Bid -> Place Another bid

## Fixed

Bid -> Cancel
Bid -> Bid Button -> Cancel
Bid -> Bid Button -> Artsy Login
Bid -> Bid Button -> Unknown Phone Number

So now it's just sign-in/bidding that's causing the problem. Will continue.

ashfurrow:
Yesterday, Andy Matuschack said something to the effect of "it's bananas that you can write a reference cycle in Swift" and I think he was right.

alloy:
@ashfurrow It’s kind of inherent with ARC style systems. You _have_ to know what you’re doing, except you don’t need to do all the actual work. You can get a lot better with a garbage collector, but even then it’s still possible. (Did you know the Ruby stdlib comes with a `WeakRef` helper class?)

On whether or not it _should_ be possible I agree, the ideal language is one where you don’t need to.

alloy:
The cycle fixes look good to me, but you should probably let someone that’s used to Swift+RAC take a look as well.

ashfurrow:
Agreed on all counts, thanks for the feedback! Andy's talk was kind of "what might swift have looked like if it didn't need ObjC interop?" A neat thought experiment.

ashfurrow:
The remaining issue is really the following cycle, which affects anything that touches loading (registration, bidding). 

> Fulfillment controller -> Loading VC -> Loading VM -> Fulfillment controller

The fulfillment controller is used primarily to access the logged-in provider. But the provider doesn't exist until the user is logged in. Quite the pickle. 

The tests will need to be updated, too.

ashfurrow:
Phew, ok. 

screen shot 2015-11-02 at 4 32 21 pm

Tests passing locally. I solved the issue mentioned above using `unowned` references, and also spotted a reference cycle in `didSet` for `xAccessToken`, so _any_ time the user was logged in, there was a leak. 

Hopefully Travis will pass. If not, well, there's a Trello card for that, too.

ashfurrow:
Yup, same problem as before. 

@orta this is good to merge now, passes locally. I'll deal with the CI issues separately so we can get this merged.

Kiosk/Auction Listings/ListingsCountdownManager.swift
```diff
@@ -21,6 +23,11 @@ class ListingsCountdownManager: NSObject {
         }
     }
 
+    /// Immediately invalidates all timers. No further updates will be made to the UI after this method is called.
```
orta: comment feels like it implies multiple timers, niggle

Kiosk/Bid Fulfillment/ConfirmYourBidArtsyLoginViewController.swift
```diff
@@ -31,21 +30,20 @@ class ConfirmYourBidArtsyLoginViewController: UIViewController {
...
-
-        let emailTextSignal = emailTextField.rac_textSignal()
-        let passwordTextSignal = passwordTextField.rac_textSignal()
-        RAC(nav.bidDetails.newUser, "email") <~ emailTextSignal.takeUntil(viewWillDisappearSignal())
-        RAC(nav.bidDetails.newUser, "password") <~ passwordTextSignal.takeUntil(viewWillDisappearSignal())
+        
+        let emailTextSignal = emailTextField.rac_textSignal().takeUntil(viewWillDisappearSignal())
+        let passwordTextSignal = passwordTextField.rac_textSignal().takeUntil(viewWillDisappearSignal())
+        RAC(nav.bidDetails.newUser, "email") <~ emailTextSignal
+        RAC(nav.bidDetails.newUser, "password") <~ passwordTextSignal
```
orta: interesting, what's going on here? ( or was this just a style choice? )

Kiosk/Bid Fulfillment/ConfirmYourBidArtsyLoginViewController.swift
```diff
@@ -31,21 +30,20 @@ class ConfirmYourBidArtsyLoginViewController: UIViewController {
...
         let inputIsEmail = emailTextSignal.map(stringIsEmailAddress)
         let passwordIsLongEnough = passwordTextSignal.map(isZeroLengthString).not()
         let formIsValid = RACSignal.combineLatest([inputIsEmail, passwordIsLongEnough]).and()
 
         confirmCredentialsButton.rac_command = RACCommand(enabled: formIsValid) { [weak self] _ in
-            if (self == nil) {
-                return RACSignal.empty()
-            }
-            return self!.xAuthSignal().`try` { (accessTokenDict, errorPointer) -> Bool in
+            guard let me = self else { return RACSignal.empty() }
```
orta: I like this

Kiosk/Bid Fulfillment/ConfirmYourBidViewController.swift
```diff
@@ -49,36 +48,33 @@ class ConfirmYourBidViewController: UIViewController {
...
+            guard let me = self else { return RACSignal.empty() }
+
+            let endpoint: ArtsyAPI = ArtsyAPI.FindBidderRegistration(auctionID: auctionID, phone: String(me.number))
+            return XAppRequest(endpoint, provider:me.provider).filterStatusCode(400).doError { (error) -> Void in
+                guard let me = self else { return }
+
+                // Due to AlamoFire restrictions we can't stop HTTP redirects
+                // so to figure out if we got 302'd we have to introspect the
+                // error to see if it's the original URL to know if the
+                // request suceedded
```
orta: typo

Kiosk/Bid Fulfillment/FulfillmentNavigationController.swift
```diff
@@ -24,12 +24,13 @@ class FulfillmentNavigationController: UINavigationController, FulfillmentContro
     /// Otherwise we're fine with a legit auth token
     var xAccessToken: String? {
         didSet(oldToken) {
+            let accessToken = self.xAccessToken!
```
orta: guard?

Kiosk/Bid Fulfillment/RegisterViewController.swift
```diff
@@ -52,17 +51,16 @@ class RegisterViewController: UIViewController {
...
         goToViewController(nextVC)
-
     }
 
     func goToViewController(controller: UIViewController) {
         self.internalNavController()!.viewControllers = [controller]
 
         if let subscribableVC = controller as? RegistrationSubController {
-            subscribableVC.finishedSignal.subscribeCompleted({ [weak self] () -> Void in
+            subscribableVC.finishedSignal.subscribeCompleted { [weak self] () -> Void in
```
orta: `Void in` can be removed

KioskTests/Bid Fulfillment/PlaceBidNetworkModelTests.swift
```diff
@@ -1,8 +1,8 @@
 import Quick
 import Nimble
 import ReactiveCocoa
-import Swift_RAC_Macros
-import Moya
+import
```
orta: this works/

orta:
Looks cool to me

Kiosk/Bid Fulfillment/ConfirmYourBidArtsyLoginViewController.swift
ashfurrow: (We talked about this in-person, summarizing here.)

The `emailTextSignal` and `passwordTextSignal` variables are used further down, but the `takeUntil` was only being applied to the first two uses of them, so I moved the call to the variable declaration.

Kiosk/Bid Fulfillment/ConfirmYourBidArtsyLoginViewController.swift
ashfurrow: Have to credit @acjay with this.

Kiosk/Bid Fulfillment/RegisterViewController.swift
ashfurrow: `in` needs there for `[weak self]`, but taking out `() -> Void` is :+1:

KioskTests/Bid Fulfillment/PlaceBidNetworkModelTests.swift
ashfurrow: Ah, yeah. Sorry, project-wide find-and-replace got messy. I was surprised, too.

ashfurrow:
No substantially changes, since CI is not reliable and I got the :+1: I'm going to merge.

Kiosk/Bid Fulfillment/BidderNetworkModel.swift
```diff
@@ -1,12 +1,11 @@
 import Foundation
 import ReactiveCocoa
-import Swift_RAC_Macros
 import Moya
 
 class BidderNetworkModel: NSObject {
     // MARK: - Getters
 
-    var fulfillmentController: FulfillmentController
+    unowned let fulfillmentController: FulfillmentController
```
acjay: I think even better might be not to inject the entire `FulfillmentController` in the first place, but instead have `FulfillmentController` inject functions that abstract over the queries / side effects the child wants. I've found it to be a really strong pattern is to never have a child component have an explicit reference to a parent. It also makes mocking much simpler -- just pass in a lamba with the assumed behavior.

ashfurrow: Makes sense – something to keep in mind when I get to https://github.com/artsy/eidolon/issues/433

PR #556, opened 2015-11-11
@orta, @acjay: Initial work moving to Swift-based FRP library

ashfurrow:
These are _very_ early stages, but I wanted to get this out there and get your reactions. 

Some notes: 

- Currently depends on [changes I've made to RxSwift](https://github.com/ReactiveX/RxSwift/compare/develop...ashfurrow:develop?expand=1). Once I finish transitioning Eidolon, there'll be some awesome pull requests. /cc @bontoJR
- Take a look at `Observable+Operators.swift` for some neat things I'm doing with the type system. Going to be more of this soon, will likely end up in either RxSwift or a library of our own. 
- I'm keeping things on a `develop` branch for now, so we can easily make changes to our deployable master. 

When either of you is cool with the changes, we can merge into the `develop` branch without CI or anything :sunglasses:


Kiosk/App/CardHandler.swift
```diff
@@ -1,9 +1,14 @@
 import UIKit
-import ReactiveCocoa
+import RxSwift
 
 class CardHandler: NSObject, CFTReaderDelegate {
 
-    let cardSwipedSignal = RACSubject()
+    private let _cardStatus = BehaviorSubject(value: "")
```
acjay: Not sure I understand the pattern here -- why does this variable have an underscore, and why does the Observable below close over it?

ashfurrow: Underscores are a pattern in iOS to indicate private variables. `_cardStatus` is private so only this class may change its `value`, even though we expose its observable publicly.

ashfurrow:
If @orta and @acjay are OK with the changes so far, I'd love to get this merged into `develop`.

Kiosk/App/CardHandler.swift
acjay: Ah, I see. Seems good to me.

I'd like to standardize a pattern for representing mutable app state (whether that's external data model or UI state). perhaps this is it!

ashfurrow: Yeah, maybe! Not entirely sold on it, but I'm taking notes.

bontoJR: Is there any reason to have a `BehaviorSubject` instead of a `PublishSubject`? 
Later in this same file the `cardStatus` is then defined as an observable with the first value skipped, so what is the point to have an initial value if this is later skipped? I am asking b/c maybe there's a technical reason for this and I am curious :)

ashfurrow: Good question! I used it because I didn't know about the other one :P Thanks for the note, I'll change :bow:

PR #558, opened 2015-11-12
@orta => Updates to latest Fastlane

ashfurrow:
Also removes the custom action, since it's not a part of the tool :tada:


orta:
:+1:

orta:
second_curtain still brings in nokogiri FWIW, which raises the barrier to entry for run-by cloners

PR #564, opened 2015-12-02
Code Golfing

ashfurrow:
Am I using that term correctly?

- Moves to trailing closure syntax where appropriate. 
- Removes "signal" nomenclature.
- Removes unnecessary closure typing.

### DIFF
--- a/Kiosk/App/Models/Artwork.swift
+++ b/Kiosk/App/Models/Artwork.swift
@@ -39,9 +39,9 @@ final class Artwork: NSObject, JSONAbleType {
     dynamic var images: [Image]?
 
     lazy var defaultImage: Image? = {
-        let defaultImages = self.images?.filter({ (image) -> Bool in
+        let defaultImages = self.images?.filter { (image) -> Bool in
             image.isDefault
-        })
+        }
 
         return defaultImages?.first ?? self.images?.first
     }()
--- a/Kiosk/App/Models/SaleArtworkViewModel.swift
+++ b/Kiosk/App/Models/SaleArtworkViewModel.swift
@@ -55,9 +55,9 @@ extension SaleArtworkViewModel {
         return saleArtwork.id
     }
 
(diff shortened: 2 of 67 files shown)


Kiosk/App/Models/Artwork.swift
```diff
@@ -39,9 +39,9 @@ final class Artwork: NSObject, JSONAbleType {
     dynamic var images: [Image]?
 
     lazy var defaultImage: Image? = {
-        let defaultImages = self.images?.filter({ (image) -> Bool in
+        let defaultImages = self.images?.filter { (image) -> Bool in
             image.isDefault
-        })
```
orta: couldn't that go into a one liner `$0.isDefault` ?

Kiosk/App/Models/SaleArtworkViewModel.swift
```diff
@@ -130,13 +130,13 @@ extension SaleArtworkViewModel {
     }
 
     func currentBidOrOpeningBid() -> Observable<String> {
-        let signals = [
+        let s = [
```
orta: `observables`?

Kiosk/Bid Fulfillment/ConfirmYourBidPINViewController.swift
```diff
@@ -5,14 +5,14 @@ import Action
 
 class ConfirmYourBidPINViewController: UIViewController {
 
-    var pin = Variable("")
+    var _pin = Variable("")
```
orta: `private var...`?

orta:
Correct use of the term. Nothing considered a blocker.

Kiosk/App/Models/SaleArtworkViewModel.swift
ashfurrow: Ha, this must've got caught up in my find/replace "signal" with "" sweep :)

PR #568, opened 2015-12-07
@orta: Fixes problems with rx_observe

ashfurrow:
_Turns out_ that observing an `NSNumber?` and `NSNumber` behave the same – RxSwift appears to unwrap the optional for us (since `rx_observer` returns an observable that _always_ sends optionals (we can't trust Objective-C)).


Podfile
```diff
@@ -59,8 +59,8 @@ pod 'XNGMarkdownParser'
 
 # Swift pods
 pod 'SwiftyJSON'
-pod 'RxSwift', :git => 'https://github.com/ReactiveX/RxSwift.git', :branch => 'develop'
-pod 'RxCocoa', :git => 'https://github.com/ReactiveX/RxSwift.git', :branch => 'develop'
+pod 'RxSwift', '~> 2.0.0-beta'
+pod 'RxCocoa', '~> 2.0.0-beta'
```
orta: interesting, I didn't know this synthx worked

orta:
Great, mergable on :green_apple:

Podfile
ashfurrow: ¯\_(ツ)_/¯

PR #571, opened 2015-12-14
Consolidate Providers with DI

### LINKED ISSUE #433: Consolidate Provider usage

ashfurrow:
This PR fixes #433 , a long-standing issue that I've been meaning to tackle for a while. Instead of using shared, global state (through `fulfillmentNav()`, a global-ish function), this uses DI to inject network access into objects.

This PR also removes some old `TODO`s and does a bit of cleanup around weak references to `self` that didn't really need to be weak.

I also tried to use the type system a bit to differentiate between regular network access and authenticated network access. Ideally, these would be stored in two separate `MoyaTarget` enums, grouped by whether or not they required authorization. It only gets hairy because the XApp token would either need to exist on both (or we assume that the XApp token never needs to be refreshed by a network call that needs authorization, a very unlikely situation). Going to take another look tomorrow.

Still to be done:

- [x] Two enums.
- [x] More manual testing
- [ ] Updated unit tests
- [ ] ~~I'd _like_ to get rid of the `fulfillmentNav()` altogether, since it's only being used to pass around the `bidDetails`, which can be similarly DI'd. But that's a lot lower priority, so maybe next time :wave:~~


ashfurrow:
Current testing status:

Registration
- By itself :(
- By itself with existing user :(
- While bidding 🎉

Bidding
- Via phone number  🎉
- Via bidder number 🎉
- Via Artsy creds :(
- Increasing bid 🎉
- From the details page 🎉

acjay:
Mammoth PR! But disentangling `FulfillmentController` is a big win. I'm a big fan of the principle of passing dependencies into a module in the absolute simplest possible form. To the extent that it's possible to depend only on types that are some closure of primitives, collections, structs, and (maybe) async containers, it pays off big time in maintainability and minimizing error surface.

ashfurrow:
I really like [this idea](https://twitter.com/jakemarsh/status/678379105658626048):

jakemarsh_2015-dec-19

But I need to get CI green first >.<

jakemarsh:
Here's the full version of the code snippet @ashfurrow pasted above:

``` swift
extension ObservableType where E == Response {
  func mapObject<T: Decodable>() -> Observable<T> {
    return flatMap { response -> Observable<T> in
      return just(try T.decode(response.mapJSON()))
    }
  }
  
  func mapObject<T: Decodable>() -> Observable<[T]> {
    return flatMap { response -> Observable<[T]> in
      return just(try [T].decode(response.mapJSON()))
    }
  }
}

let disposeBag = DisposeBag()
let RedditAPI = RxMoyaProvider<Reddit>()

// Reddit is the TargetType
// RedditAPI is the global provider instance

extension Reddit {
  func asObservable<T: Decodable>(type: T.Type, provider: RxMoyaProvider<Reddit> = RedditAPI) -> Observable<T> {
    return provider.request(self).mapObject() as Observable<T>
  }
}

func blah() {
  disposeBag.addDisposable(
    Reddit.SubredditInfo("starwars").asObservable(Subreddit).subscribe {
      switch $0 {
        case .Next(let subreddit): print(subreddit.title)
        case .Error(let error): print(error)
        default: break
      }
    }
  )
}
```

ashfurrow:
Cool, this is good to go :+1: 

I'm updating to the 2.0 stable release of RxSwift; it's non-trivial so I'm going to submit a separate PR once this one is merged.

ashfurrow:
The updates are done on another branch. See the diff at https://github.com/artsy/eidolon/compare/consolidate...updates . It updates to new RxSwift stuff and uses Moya's new error-handling mechanism to be more precise about "the bid request you submitted was invalid because the amount was too low, even if it was valid when you started to place your bid" problem.

orta:
This feels alright to me. However, in order to review this I had to: remember how Eidolon works, try think functionally, study how RxSwift works

So it'd be good to have @acjay also give a :+1:

Kiosk/Bid Fulfillment/RegistrationPasswordViewController.swift
```diff
@@ -4,12 +4,16 @@ import Moya
...
 class RegistrationPasswordViewController: UIViewController, RegistrationSubController {
+    
     @IBOutlet var passwordTextField: TextField!
     @IBOutlet var confirmButton: ActionButton!
     @IBOutlet var subtitleLabel: UILabel!
     @IBOutlet var forgotPasswordButton: UIButton!
+
     let finished = PublishSubject<Void>()
 
+    var provider: Networking!
```
acjay: How do you feel about the idea of documenting `var` any time it's used? Looking at how `RegistrationPasswordViewController` is instantiated from a storyboard, it appears that this is a `var` just because of the need for late-initialization, but not because it's truly going to vary over the lifetime of the object. I think that's a legit use. Perhaps we can just annotate it with a `// for late initialization, not mutation`.

acjay:
This looks like a big step forward! :+1:

Kiosk/Bid Fulfillment/RegistrationPasswordViewController.swift
ashfurrow: I think _every_ use of `var` fits under that umbrella :smile_cat: Unless something is marked as `dynamic`.

You're totally right about why its used – not sure what the best approach for documenting it would be though.

ashfurrow:
Cool, thanks! Comfortable hitting the 'merge' button?

Issue #576, opened 2016-01-18
bundle install don't work, how to fix on this problem ? 


uthens:
Sample: https://github.com/artsy/eidolon
----------------------------------------------------------------
git clone https://github.com/artsy/eidolon.git
cd eidolon
bundle install
bundle exec fastlane oss
----------------------------------------------------------------

>> Uthens-MacBook-Pro:eidolon uthen$ bundle install 
don't work how to resolve on this problem ? 

Last login: Mon Jan 18 22:56:11 on ttys004
Uthens-MacBook-Pro:eidolon uthen$ bundle install
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Fetching https://github.com/ffi/ffi.git
fatal: Could not parse object 'd94b923c1f574ff63743f0a3eefc917da03386b9'.
Git error: command `git reset --hard d94b923c1f574ff63743f0a3eefc917da03386b9` in directory
/Users/uthen/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/bundler/gems/ffi-d94b923c1f57 has failed.
If this error persists you could try removing the cache directory
'/Users/uthen/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/cache/bundler/git/ffi-7fa1fad54f97b3471351dbde91d8b0a9444ce49e'
Uthens-MacBook-Pro:eidolon uthen$

orta:
Did you try deleting `Users/uthen/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/cache/bundler/git/ffi-7fa1fad54f97b3471351dbde91d8b0a9444ce49e` like it asked, and re-runnning?

ashfurrow:
We're seeing this on our [Continuous Integration](https://travis-ci.org/artsy/eidolon), too. Started over the weekend, still looking into it.

ashfurrow:
Probably related to https://github.com/ffi/ffi/issues/461#issuecomment-172568651

uthens:
Did you try deleting Users/uthen/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/cache/bundler/git/ffi-7fa1fad54f97b3471351dbde91d8b0a9444ce49e like it asked, and re-runnning? 

<<< I try to deleted on folder and re-rerunning, but still have the same problem..

[...]

PR #577, opened 2016-01-18
[CI] Removes bundler cache.

ashfurrow:
Possibly fixes CI. We'll see!

Related: https://github.com/artsy/eidolon/issues/576


ashfurrow:
Nope :disappointed:

PR #582, opened 2016-01-27
Fix crash in Swift 2.2 compiler with Xcode 7.3

johnno1962:
Hi, In this code under SaleArtworkDetailsViewController.setupAdditionalDetailStackView:

```
if let artist = artist() {
            retrieveArtistBlurb()
                .filter { blurb in
                    return blurb.isNotEmpty
                }
                .subscribeNext { [weak self] blurb in
                    guard let me = self else { return }


                    let aboutArtistHeaderLabel = label(.Header)
```

The `subscribeNext` closure weakly references `self`, but also refers to the function `label` that strongly references `self` and causes Swift 2.2 to crash.

Other changes are just to get a clean build from zero. Not sure about ORStackView version required as I get a single large “help” button over the display.

### DIFF
--- a/Kiosk/Bid Fulfillment/ManualCreditCardInputViewModel.swift
+++ b/Kiosk/Bid Fulfillment/ManualCreditCardInputViewModel.swift
@@ -83,10 +83,10 @@ class ManualCreditCardInputViewModel: NSObject {
 
         return stripeManager.registerCard(cardFullDigits.value, month: month, year: year, securityCode: securityCode.value, postalCode: billingZip.value).doOnNext { token in
 
-            newUser.creditCardName.value = token.card.name
-            newUser.creditCardType.value = token.card.brand.name
+            newUser.creditCardName.value = token.card!.name
+            newUser.creditCardType.value = token.card!.brand.name
             newUser.creditCardToken.value = token.tokenId
-            newUser.creditCardDigit.value = token.card.last4
+            newUser.creditCardDigit.value = token.card!.last4()
         }
     }
 
--- a/Kiosk/Sale Artwork Details/SaleArtworkDetailsViewController.swift
+++ b/Kiosk/Sale Artwork Details/SaleArtworkDetailsViewController.swift
@@ -332,37 +332,37 @@ class SaleArtworkDetailsViewController: UIViewController {
         }
     }
 
+    private enum LabelType {
+        case Header
+        case Body
+    }
+
+    private func label(type: LabelType, layout: Observable<Void>? = nil) -> UILabel {
+        let (label, fontSize) = { () -> (UILabel, CGFloat) in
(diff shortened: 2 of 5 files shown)


ashfurrow:
Cool, thanks! Let me take a look...

Kiosk/Bid Fulfillment/ManualCreditCardInputViewModel.swift
```diff
@@ -83,10 +83,10 @@ class ManualCreditCardInputViewModel: NSObject {
 
         return stripeManager.registerCard(cardFullDigits.value, month: month, year: year, securityCode: securityCode.value, postalCode: billingZip.value).doOnNext { token in
 
-            newUser.creditCardName.value = token.card.name
-            newUser.creditCardType.value = token.card.brand.name
+            newUser.creditCardName.value = token.card!.name
+            newUser.creditCardType.value = token.card!.brand.name
             newUser.creditCardToken.value = token.tokenId
-            newUser.creditCardDigit.value = token.card.last4
+            newUser.creditCardDigit.value = token.card!.last4()
```
ashfurrow: Hmm, these are fairly large semantic changes, I'll have to take a closer look later about them.

Kiosk/Sale Artwork Details/SaleArtworkDetailsViewController.swift
```diff
@@ -401,11 +401,11 @@ class SaleArtworkDetailsViewController: UIViewController {
                 .subscribeNext { [weak self] blurb in
                     guard let me = self else { return }
 
-                    let aboutArtistHeaderLabel = label(.Header)
+                    let aboutArtistHeaderLabel = me.label(.Header)
                     aboutArtistHeaderLabel.text = "About \(artist.name)"
                     me.additionalDetailScrollView.stackView.addSubview(aboutArtistHeaderLabel, withTopMargin: "22", sideMargin: "40")
 
-                    let aboutAristLabel = label(.Body, layout: me.layoutSubviews)
+                    let aboutAristLabel = me.label(.Body, layout: me.layoutSubviews)
```
ashfurrow: Ah, I see... interesting. I would have thought the compiler would've caught those.

ashfurrow:
Cool, I like the direction :+1: It'll be next week probably before I'm back on this to more thoroughly review, but I really appreciate it!

johnno1962:
np, the weak self fix came from an Apple Bug report I filled on Xcode 7.3b2. Unsure whats up with ORStackView and Stripe versions but I’ve always had to make these changes when I clone eidonon.

ashfurrow:
Super-interesting! I'll investigate and let you know – thanks for the clarification!

ashfurrow:
Closing in lieu of #617.

PR #584, opened 2016-02-05
Fixes problem authenticating with bidder number / phone number

ashfurrow:
The [issue](https://trello.com/c/6IZJd0oU/252-showstopper-kiosk-users-who-enter-bidder-number-should-be-asked-for-pin-and-not-email-address) was caused by a change to Moya's change in error-handling that I missed.

I took a look around for other places using the old logic, found one, but it was only related to logging. 

Oh, also added an admin button to bid the next increment of a sale artwork:

screen shot 2016-02-04 at 8 45 36 pm

Should speed up debugging :smiley_cat:


Kiosk/App/SwiftExtensions.swift
```diff
@@ -54,3 +54,19 @@ extension Optional where Wrapped: Occupiable {
         return !isNilOrEmpty
     }
 }
+
+// TODO: PR this into Moya
```
ashfurrow: I have so PR'd: https://github.com/Moya/Moya/pull/401

However, we're a major version of Moya behind, and I don't have time to update right now. We'll put this on the back-burner for now.

ashfurrow:
Nice! I was hoping that would fix the CI :joy_cat:

Kiosk/App/SwiftExtensions.swift
acjay: Amused that we're a version behind on the library you started :) :)

Kiosk/Bid Fulfillment/ConfirmYourBidViewController.swift
```diff
@@ -83,7 +83,11 @@ class ConfirmYourBidViewController: UIViewController {
                     // error to see if it's the original URL to know if the
                     // request succeeded.
 
-                    let response = (error as NSError).userInfo["data"] as? Moya.Response
+                    var response: Moya.Response?
```
orta: if it's lazy set once I think you can use a `let` here.

orta:
Cool, looks good.

Kiosk/Bid Fulfillment/ConfirmYourBidViewController.swift
ashfurrow: Only if all code paths set it exactly once, I'm afraid.

ashfurrow: (We might not set it at all, in this case.)

PR #585, opened 2016-02-05
Updates Nimble+Quick

ashfurrow:
Nimble had an issue with intermittent failures, like this one: https://travis-ci.org/artsy/eidolon/builds/107248693

Updating to 3.1.0 should fix it. Also, fastlane did something to its readme ¯\\_(ツ)_/¯


orta:
Good on Green :+1:

ashfurrow:
Failure appears legit? Looking into it.

ashfurrow:
That should do it, not sure what was going on. Probably still related to threading, need to go through and take a look at all the places we're scheduling things and make sure to un-schedule them appropriately (in the app, these are often only scheduled once per lifecycle, but they mess up the tests).

mennenia:
test

PR #586, opened 2016-02-05
Fixes bid resolution problems

ashfurrow:
This is a fix for a problem tracked in Trello: https://trello.com/c/Bio2jZQe/297-kiosk-showstopper-bid-polling-for-resolving-high-bidder-status-failing#

The issue is that bids weren't being placed (unless you were also registering, strangely). Actually this makes sense, once the underlying issue was found. 

We were sending `empty()` as a result of "we don't need to do this piece of work", and then trying to `flatMap` that empty observable to the bid-placing network call. But empty observables don't have anything to `flatMap` against, they're empty. So instead, now we're sending `just()`, which is short for `just(Void())`, which is a value (an empty one). So that value can get `flatMap`'d appropriately.

tl;dr we were returning an empty observable instead of an observable that sends an empty value.

Also, I fixed a problem where if you placed a bid, but it wasn't high enough, and then you hit "place higher bid", then changed your mind, the cancel button was gone and the app wouldn't let you leave without placing a higher bid.

### DIFF
--- a/Kiosk/Bid Fulfillment/BidderNetworkModel.swift
+++ b/Kiosk/Bid Fulfillment/BidderNetworkModel.swift
@@ -100,7 +100,7 @@ private extension BidderNetworkModel {
         // If the user was asked to swipe a card, we'd have stored the token. 
         // If the token is not there, then the user must already have one on file. So we can skip this step.
         guard let token = bidDetails.newUser.creditCardToken.value else {
-            return .empty()
+            return .just()
         }
 
         let swiped = bidDetails.newUser.swipedCreditCard
@@ -110,7 +110,7 @@ private extension BidderNetworkModel {
             .filterSuccessfulStatusCodes()
             .map(void)
             .doOnCompleted { [weak self] in
-                // Adding the credit card succeeded, so we shoudl clear the newUser.creditCardToken so that we don't
+                // Adding the credit card succeeded, so we should clear the newUser.creditCardToken so that we don't
                 // inadvertently try to re-add their card token if they need to increase their bid.
 
                 self?.bidDetails.newUser.creditCardToken.value = nil
@@ -125,7 +125,7 @@ private extension BidderNetworkModel {
 
         return bool.flatMap { exists -> Observable<Void> in
             if exists {
-                return .empty()
+                return .just()
             } else {
                 return self.registerToAuction(self.bidDetails.auctionID, provider: provider).then { [weak self] in self?.generateAPIN(provider) }
             }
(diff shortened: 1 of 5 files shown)


Kiosk/Bid Fulfillment/LoadingViewController.swift
```diff
@@ -142,6 +142,7 @@ extension LoadingViewController {
 
         let title = reserveNotMet ? "NO, THANKS" : (createdNewBidder ? "CONTINUE" : "BACK TO AUCTION")
         backToAuctionButton.setTitle(title, forState: .Normal)
+        fulfillmentContainer()?.cancelButton.hidden = false
```
acjay: In a perfect world, would this attribute be hooked directly up to a signal?

Kiosk/Bid Fulfillment/BidderNetworkModel.swift
```diff
@@ -110,7 +110,7 @@ private extension BidderNetworkModel {
             .filterSuccessfulStatusCodes()
             .map(void)
             .doOnCompleted { [weak self] in
-                // Adding the credit card succeeded, so we shoudl clear the newUser.creditCardToken so that we don't
+                // Adding the credit card succeeded, so we should clear the newUser.creditCardToken so that we don't
                 // inadvertently try to re-add their card token if they need to increase their bid.
 
                 self?.bidDetails.newUser.creditCardToken.value = nil
```
acjay: I think this is one area where bugs tend to creep in -- we have mixes of FRP and of shared mutable state, which I think has the potential of dropping us to the lowest common denominator of reliability.

But not sure that's actionable right now, other than as a note to ourselves.

Kiosk/Bid Fulfillment/LoadingViewController.swift
ashfurrow: Yes, the whole `fulfillmentContainer()` is a poor abstraction that I've been meaning to replace when I get a chance. This added line is really fulfilling the promise in this comment: https://github.com/artsy/eidolon/blob/183210ea6c470e5819118ac0fda2d61fe5bc8a05/Kiosk/Bid%20Fulfillment/LoadingViewController.swift#L61

Kiosk/Bid Fulfillment/BidderNetworkModel.swift
ashfurrow: Agreed, something to be aware of as we move forward.

acjay:
It looks good to go to me.

Regarding the original problem, it does seem to me that there's a semantic gap between the observable API and the UI/data flow. We may want to find ways to abstract over the subtleties of Rx, which may help prevent logic errors.

As alluded to in the comments, long term, I think we can improve upon a couple things architectural to go all in on FRP, but probably best to separate that from fixing the problem at hand.

Good fix!

ashfurrow:
Thanks!

acjay:
I'm still seeing Place Higher Bid when the Bid Submitted screen pops up. I think the polling may still be failing.

ashfurrow:
It says the words "Bid Submitted"? Does it take a few seconds or does it happen immediately? I'll look into it.

acjay:
It's after a delay, so I think it's timing out on the polling

ashfurrow:
Hmm. Thanks.

PR #588, opened 2016-02-11
Fixes created-by-admin flag not being respected

ashfurrow:
Full discussion on Slack: https://artsy.slack.com/archives/auctions-dev/p1455065372000615

Relates to this Trello card: https://trello.com/c/Bio2jZQe/297-kiosk-showstopper-bid-polling-for-resolving-high-bidder-status-failing

tl;dr Gravity exposes a `created_by_admin` field for the bidder model that the site uses to bypass the requirement that a bidder have a credit card. On the Kiosk, the app would get into an unexpected state (one where the user was logged in but asked to go through the regular CC-collection flow). That led them to the password prompt, which is confusing.

We've had this bug for _months_ and this should finally squash it! Thanks to Barry for his work on getting the change in Gravity done so quickly: https://github.com/artsy/gravity/pull/9705

I will comment inline with specifics.


Kiosk/Bid Fulfillment/ConfirmYourBidPINViewController.swift
```diff
@@ -65,15 +66,20 @@ class ConfirmYourBidPINViewController: UIViewController {
                         .updateUserCredentials(loggedInProvider)
                         .mapReplace(provider)
                 }
-                .flatMap { provider -> Observable<AuthorizedNetworking> in
+                .flatMap { provider -> Observable<BypassResults> in
```
ashfurrow: This is the big addition, but it doesn't _look_ like an addition. That's because we were calling `updateUserCredentials` twice for no reason, and I removed it in https://github.com/artsy/eidolon/commit/a096ff10cff5ee6e300598051381fac5f521c20c

Logic for making the network call and determining if CC bypass should be made is contained completely within the network model. Returning an Observable of a tuple is a new thing I'm trying, not sure if I like it or not.

Kiosk/Bid Fulfillment/AdminCCBypassNetworkModel.swift
```diff
@@ -0,0 +1,37 @@
...
+
+protocol AdminCCBypassNetworkModelType {
+
+    func checkForAdminCCBypass(saleID: String, authorizedNetworking: AuthorizedNetworking) -> Observable<BypassResults>
+}
+
+class AdminCCBypassNetworkModel: AdminCCBypassNetworkModelType {
+
+    /// Returns an Observable of (Bool, AuthorizedNetworking)
+    /// The Bool represents if a the Credit Card requirement should be waived.
```
orta: > if a the Credit Card

Kiosk/Bid Fulfillment/ConfirmYourBidPINViewController.swift
orta: It does feel a bit weird, but nowhere close to being strongly enough that I'm adverse

Kiosk/Bid Fulfillment/ConfirmYourBidPINViewController.swift
```diff
@@ -65,15 +66,20 @@ class ConfirmYourBidPINViewController: UIViewController {
...
-                        .fulfillmentNav()
-                        .updateUserCredentials(loggedInProvider)
-                        .mapReplace(provider)
+                        .viewModel
+                        .checkForAdminCCBypass(bidDetails.auctionID, authorizedNetworking: provider)
                 }
-                .flatMap { provider -> Observable<AuthorizedNetworking> in
+                .flatMap { results -> Observable<Void> in
+                    // Check if we should bypass the CC requirement and move directly onto the next step.
+                    guard results.bypassCCRequirement == false else {
```
orta: I feel like the guard here kinda makes the logic a bit harder to follow, might be easier to just do an if

Kiosk/Bid Fulfillment/ConfirmYourBidPINViewController.swift
```diff
@@ -83,9 +89,8 @@ class ConfirmYourBidPINViewController: UIViewController {
                                 me.performSegue(.PINConfirmedhasCard)
                             }
                         }
-                        .mapReplace(provider)
+                        .map(void)
```
orta: Giphy

Kiosk/Bid Fulfillment/ConfirmYourBidPINViewController.swift
ashfurrow: Yeah, feels weird.

--

[A.]           Ash Furrow

Artsy <https://www.artsy.net/>

                *ash@artsy.net <ash@artsy.net>*

                (347)-592-4474 <//(347)-592-4474> (Mobile)
                @ashfurrow <http://twitter.com/ashfurrow>

Kiosk/Bid Fulfillment/AdminCCBypassNetworkModel.swift
```diff
@@ -0,0 +1,37 @@
...
+protocol AdminCCBypassNetworkModelType {
+
+    func checkForAdminCCBypass(saleID: String, authorizedNetworking: AuthorizedNetworking) -> Observable<BypassResults>
+}
+
+class AdminCCBypassNetworkModel: AdminCCBypassNetworkModelType {
+
+    /// Returns an Observable of (Bool, AuthorizedNetworking)
+    /// The Bool represents if a the Credit Card requirement should be waived. 
+    /// THe AuthorizedNetworking is the same instance that's passed in, which is a convenience for chaining observables.
```
orta: > THe AuthorizedNetworking

orta:
Implementation feels fine, if you're happy with the code - that's fine, otherwise refactor and you're good to go :+1:

Kiosk/Bid Fulfillment/ConfirmYourBidPINViewController.swift
ashfurrow: I've been looking for an alternative to how we do this now with `.mapReplace(provider)`, and this isn't it. I'd rather keep this consistent with what we do now.

Kiosk/Bid Fulfillment/ConfirmYourBidPINViewController.swift
ashfurrow: This feels weird because I should've been using an enum with explicit cases instead of true/false. I think I was too distracted trying to get an improvement to the `.mapReplace(provider)` pattern :wink: Definitely merits more thought.

ashfurrow:
Yeah, the two-case-enum-instead-of-Bool idea might warrant a blog post 🤔

PR #592, opened 2016-03-01
Use cell re-use based disposal bags for bindings on cells

### LINKED ISSUE #591: Refreshing networking data sends the wrong data to existing cells
It's looking like we send the wrong starting/current bid data to the wrong cell during a sync of the sale artworks. It's easy to re-pro:

orta:
Instead of de-init based ones - fixes #591

Gonna take a quick stab at  tests now, if in a few hours there are none, that's probably cause I moved back to eigen auctions work.


ashfurrow:
Interesting! The fix works for sure, but I'm curious why the problem is there in the first place. I may take a look at the `flatMapTo` (which is of my own invention, so).

orta:
:+1:

acjay:
:raised_hands:

PR #605, opened 2016-07-15
Changes typealias to associatedtype inside protocols

pedrovereza:
`typealias` usage inside protocols is deprecated in Swift 2.2 and will be removed in Swift 3 according to [this proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0011-replace-typealias-associated.md). `associatedtype` should be use instead.


ashfurrow:
The CI failure isn't you, thanks for the PR!

PR #606, opened 2016-07-17
Fix deprecations

KVTaniguchi:
The purpose of this pull request is to fix some of the deprecation warnings.  Did not fix the curried deprecation warnings as some of the observables seem to be dependent on those functions.


ashfurrow:
Awesome, thanks!

PR #613, opened 2016-09-08
Fix for tall images

### LINKED ISSUE #612: Tall images are blurry
On the latest auction.

ashfurrow:
Fixes #612.

### DIFF
--- a/Kiosk/App/Models/Image.swift
+++ b/Kiosk/App/Models/Image.swift
@@ -60,10 +60,9 @@ final class Image: NSObject, JSONAbleType {
 
     func thumbnailURL() -> NSURL? {
         let preferredVersions = { () -> Array<String> in
-            // This is a hack for https://www.artsy.net/artwork/d-star-face-work-on-paper-number-5
-            // It's a very tall image and the "medium" version looks terribad.
+            // For very tall images, the "medium" version looks terribad.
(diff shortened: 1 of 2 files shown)


Kiosk/App/Models/Image.swift
```diff
@@ -60,10 +60,9 @@ final class Image: NSObject, JSONAbleType {
 
     func thumbnailURL() -> NSURL? {
         let preferredVersions = { () -> Array<String> in
-            // This is a hack for https://www.artsy.net/artwork/d-star-face-work-on-paper-number-5
-            // It's a very tall image and the "medium" version looks terribad.
+            // For very tall images, the "medium" version looks terribad.
             // In the long-term, we have an issue to fix this for good: https://github.com/artsy/eidolon/issues/396
-            if self.id == "56ba2884139b211c61000204" {
+            if ["57be35d7a09a6711ab004fa5", "57be1fb4cd530e65fe000862"].contains(self.id) {
```
orta: haha - could you look at the aspect ratio?

ashfurrow: That seems like a better idea, though I'm wary of showing too many large/larger images on screen at once. I think this solution of specific IDs works for the short-term until we implement the grid design in #396.

orta:
👍

PR #616, opened 2016-09-30
Use TestAppDelegate to prevent normal app launch during testing

scottrhoyt:
When I was going through the Eidolon code looking for examples of Moya usage, I noticed the work around for preventing normal app launch while testing. I thought I would share the approach that I have used before. It's more code and more files, but I like that it gets any reference to `XCTest` out of `AppDelegate` and prevents `AppDelegate` from ever loading at all--so less risk of unintended side effects during testing. For these reasons, I think it is slightly more elegant. Unfortunately I was coding blind on this one because of Xcode 8 and a problem with `cocoapods-keys`, so I'm not 100% sure it even compiles 😬. But the idea is there if you want to use it at all. If not, no worries, the current solution is certainly more concise. Cheers!

### DIFF
--- a/Kiosk.xcodeproj/project.pbxproj
+++ b/Kiosk.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		3B9F6AC41D9E230F00A44C26 /* main.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3B9F6AC31D9E230F00A44C26 /* main.swift */; };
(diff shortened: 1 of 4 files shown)


orta:
Ah yeah! We use this technique in Eigen and Energy - I'm totally cool wit this PR if we get a human to verify on our side 👍 

Thanks @scottrhoyt - nice contribution

scottrhoyt:
No problem!

ashfurrow:
Looks good, I'll verify locally on Monday. Thanks!

ashfurrow:
Looks like eidolon actually uses quite a bit of the app delegate, even under testing, so we actually need to keep it as-is. I think our existing workaround (to isolate side effects from tests) is okay, not great, but it works :smile: Thanks again for the PR!

PR #626, opened 2016-11-17
Update README getting started to Swift 3

BasThomas:
The `README` still stated you need to use `Xcode 7`, as the project is Swift 2. As the project is now using Swift 3, this should be `Xcode 8`.


orta:
👍 shame about the CI though

PR #627, opened 2016-11-17
Fixes transitive dependencies for OSS font pods.

### LINKED ISSUE #625: CocoaPods keep fetching Artsy+UIFonts
I know that in your team you guys use Artsy_UIFonts module, while we use Artsy_OSSUIFonts module when building the oss version.
screen shot 2016-11-16 at 1 48 45 pm
Yet when i execute "fastlane oss" or "pod install", cocoapods keep fetching the Artsy+UIFonts module. Is there a cache somewhere. I've tried to delete Pod.lock, not working?
screen shot 2016-11-16 at 1 48 10 pm

ashfurrow:
Fixes #625.


ashfurrow:
Mmmm. Test results at https://eidolon-ci.s3.amazonaws.com/176828216/index.html suggest that we're now using new fonts, may need snapshot updates. At first glance I don't see a problem with it. 

@katarinabatina any objections to using the new iOS fonts for the Kiosk? 

Before:

reference_when_displaying_stubbed_contents__without_lot_numbers__a_listings_controller__least_bids 2x


After:

failed_when_displaying_stubbed_contents__without_lot_numbers__a_listings_controller__least_bids 2x

orta:
I can't really tell the difference :)

ashfurrow:
Looks like it's mainly (only?) the kerning on the sans serif font:

screen shot 2016-11-19 at 10 36 58 am

ashfurrow:
@orta could you review the changes I made to CI config? They're in this range: https://github.com/artsy/eidolon/pull/627/files/1641c47065771c86c6b47db5a57d961b64c57f87..9c5178ea1eabd6e4644ec322bb88c06e2a1f4359

circle.yml
```diff
@@ -1,13 +1,24 @@
...
+    version: "8.1.0"
+  environment:
+    FASTLANE_SKIP_UPDATE_CHECK: true
+    BUNDLE_PATH: vendor/bundle
+    UPLOAD_IOS_SNAPSHOT_BUCKET_NAME: eidolon-ci
 dependencies:
+  cache_directories:
+    - vendor/bundle
+    - ~/.cocoapods/repos
   override:
```
orta: I recommend you co not override dependencies

orta:
Alright, so I'd recommend taking a second shot at the circle - use [this](https://github.com/artsy/eigen/blob/7ee267cdce7d23d5bc5ca9f21302f4faa042d8ac/circle.yml) file as your reference - you should get much faster builds

orta:
The key advantages:

* The default CircleCI dependencies is 👍 (better than ours)
* If we can _build_ the app in the dependencies `post` than we can also cache the derived data, but you need to add a known derived data folder to the `xcodebuild` args

circle.yml
ashfurrow: Thanks for the tips, going to be tricky not to override dependencies since we rely on `fastlane` to set up CocoaPods-Keys keys. I'll read up on Circle's docs 👍

orta: That's a trade-off, you're welcome to make that call 👍

orta: @alloy I got a `pod install` down to a very small amount of time on CI now using these techniques ^=

https://circleci.com/gh/artsy/eigen/3539?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
screen shot 2016-11-22 at 19 28 23

alloy: That’s def ace 👌 but I’m less worried about the time over just installation working at all.

katarinabatina:
sorry for the late response yes i think this looks good

circle.yml
orta: hah, fair

PR #632, opened 2016-12-06
Makes fulfillment UI conform to Artsy fonts.

### LINKED ISSUE #630: Storyboard fonts need to be set in code
It appears to be using the system font instead of our custom fonts:

img_0052

This bug is only present on the iOS 10 release.

ashfurrow:
Fixes #630.


orta:
cool - looks fine to me 👍

PR #633, opened 2016-12-08
update .ruby-version to 2.3.0

AndrewSB:
Thought I'd go ahead and update this to the latest ruby. I tried a both `bundle install` and `bundle exec fastlane oss`, and it seemed to work for me.


ashfurrow:
Cool, thanks 👍

PR #634, opened 2016-12-09
[WIP] integrate an external networking stack

AndrewSB:
Hey @ashfurrow!

A little embarrassing to admit, but I was having trouble setting up the repo with cocoapods. I've never really used cocoapods before. When you have some time, could you point me in the right direction with https://github.com/AndrewSB/Networking/issues/1?

### DIFF
--- a/Kiosk/App/Networking/Networking.swift
+++ b/Kiosk/App/Networking/Networking.swift
@@ -3,39 +3,6 @@ import Moya
 import RxSwift
 import Alamofire
 
-class OnlineProvider<Target>: RxMoyaProvider<Target> where Target: TargetType {
-
-    fileprivate let online: Observable<Bool>
-
-    init(endpointClosure: @escaping EndpointClosure = MoyaProvider.DefaultEndpointMapping,
-        requestClosure: @escaping RequestClosure = MoyaProvider.DefaultRequestMapping,
-        stubClosure: @escaping StubClosure = MoyaProvider.NeverStub,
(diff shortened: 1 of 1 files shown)


orta:
I'm a bit wary of adding a new dependency that is 3 days old and doesn't have a README...

Especially one where we would end up having to add CocoaPods support - was there some discussion around this that I've not seen?

ashfurrow:
Yeah, I asked for a PR to discuss what could be abstracted from Eidolon for the purposes of illustrating what Moya and RxSwift could do together.

ashfurrow:
@AndrewSB I can't remember where we landed on this, should we keep this open or close it?

AndrewSB:
I'm not sure, I can't remember either. I'll leave it up to you 👍

I think something separate might be easier to digest for people like the author of https://github.com/Moya/Moya/issues/936
But I think the separate repo should be under Moya, and not my user, IF we are to make it.

pedrovereza:
@ashfurrow @AndrewSB I think you both agreed that extracting these things to a repo was an overkill (see [here](https://github.com/Moya/Moya/issues/722#issuecomment-266578494)).

I feel the overall conclusion was that having something like [AndrewSB/Networking](https://github.com/AndrewSB/Networking) is useful for Moya contributors to team up and solve common use cases and use the repo as reference for good practices.

However, I think the idea of integrating it with Eidolon was discarded (too small, overkill, etc).

ashfurrow:
Cool, gonna close then, thanks!

PR #650, opened 2017-02-06
Update License section

jonallured:
I noticed that our copyright line was inconsistent across projects and
often still listed 2016. So, I worked with Yayoi to figure out what the
right language was and am now copy/pasting to achieve consistency.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,13 @@
-The MIT License (MIT)
+MIT License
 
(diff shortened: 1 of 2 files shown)


jonallured:
Ok, last note on license/copyright stuff today, promise! Turns out the copyright line belongs in the LICENSE file and that should be the single source of truth for this info. While I was at it, I updated the copy slightly to match the default copy from the GitHub template.

PR #657, opened 2017-04-10
Filter sold artworks

ashfurrow:
Most of the plumbing was already in place, but wasn't getting used because of stringly-typed code. Snapshots updated due to antialiasing differences. 

This is a WIP because I want to test in an environment where some sale artworks get sold and some are still for sale.


ashfurrow:
Confirmed this works. The Kiosks sync every sixty seconds, so there will be a slight lag. Still remains to be seen if we need something for users already in the bidding flow. 

simulator screen shot apr 10 2017 2 14 33 pm
simulator screen shot apr 10 2017 2 14 36 pm

ashfurrow:
Failures are at https://eidolon-ci.s3.amazonaws.com/207/index.html , looks like more antialiasing problems. Will take a look later today.

orta:
Yep, this looks 👍

Issue #658, opened 2017-04-14
App unable to connect to newer card readers


alessandra-castillo:
When in "Card Tester" mode with the newer card swipers, we only see: 

_Started searching
Reader is attached_

It does not reach _Reader is connected_ so we can't use the new swipers for capturing CC's right now.

alessandra-castillo:
Hey @ashfurrow, did you hear back from Cardflight about this issue? It would be great to make use of the new swipers for our PP event on May 2, since 2 of our kiosks will be in LA at the same time. cc: @melanieborinstein

ashfurrow:
I've not heard back from them, I'll poke them.

PR #660, opened 2017-04-14
Defaults to unknown reader type, logging improvements.

ashfurrow:
Newer card readers weren't "connecting", so this changes it and adds a few other improvements. They're now reading the card, but not able to tokenize them. I've opened [an issue](https://github.com/CardFlight/cardflight-ios/issues/72) to ask for help.

Fixes https://github.com/artsy/eidolon/issues/658

### DIFF
--- a/Kiosk/Admin/AdminCardTestingViewController.swift
+++ b/Kiosk/Admin/AdminCardTestingViewController.swift
@@ -38,7 +38,7 @@ class AdminCardTestingViewController: UIViewController {
                         return
                     }
 
-                    let cardDetails = "Card: \(card.name) - \(card.last4) \n \(card.cardToken)"
+                    let cardDetails = "Card: \(card.name ?? "") - \(card.last4 ?? "") \n \(card.cardToken ?? "")"
                     self.log(cardDetails)
                 }
             }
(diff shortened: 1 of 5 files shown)


Kiosk/Admin/AdminCardTestingViewController.swift
```diff
@@ -38,7 +38,7 @@ class AdminCardTestingViewController: UIViewController {
                         return
                     }
 
-                    let cardDetails = "Card: \(card.name) - \(card.last4) \n \(card.cardToken)"
+                    let cardDetails = "Card: \(card.name ?? "") - \(card.last4 ?? "") \n \(card.cardToken ?? "")"
```
orta: Might be worth adding p`"[unknown]` instead?

ashfurrow: The card should always be there, but it's an IUO and I just added `?? ""` to avoid SwiftLint warnings :)

orta:
Heh, the fix is to go back a bunch of versions - ouch

ashfurrow:
Yeah, they're working on a better solution but yeah, seems like the bug has been present for a while.

orta:
Then if you're happy, I'm a 👍

PR #661, opened 2017-04-19
Various Fixes

ashfurrow:
Three main fixes:

- We now list all (up to a hundred) sales for admins to switch between; before it was capped to the default page size of ten.
- The symbol field was using the system default font, I changed it to Artsy's sans serif font and repositioned it to match.
- Fixes the conditional cast (`as? Currency` vs `?.currencyValue`). This was causing a problem ([reported here](https://artsy.slack.com/archives/C0C4AJ1PF/p1492617018164900)) because the cast would always fail.

CI has been giving us issues, may not pass the first time.


Kiosk/Storyboards/Fulfillment.storyboard
```diff
@@ -89,7 +89,7 @@
         <scene sceneID="wG8-BK-Gy6">
             <objects>
                 <navigationController modalPresentationStyle="formSheet" navigationBarHidden="YES" id="w8O-U4-lhE" customClass="FulfillmentNavigationController" customModule="Kiosk" customModuleProvider="target" sceneMemberID="viewController">
-                    <navigationBar key="navigationBar" contentMode="scaleToFill" misplaced="YES" id="PvK-mZ-4Jp">
+                    <navigationBar key="navigationBar" contentMode="scaleToFill" id="PvK-mZ-4Jp">
```
erikdstock: What is this?

Kiosk/Storyboards/Fulfillment.storyboard
```diff
@@ -315,7 +315,7 @@
                             <constraint firstItem="ic1-FX-iJp" firstAttribute="leading" secondItem="4Nc-B4-Sow" secondAttribute="leading" id="9V5-lV-cUq"/>
                             <constraint firstItem="UKQ-z8-PIR" firstAttribute="top" secondItem="GS2-EK-4Nx" secondAttribute="bottom" constant="10" id="Bb8-ER-cpC"/>
                             <constraint firstItem="sft-xL-TDA" firstAttribute="top" secondItem="Y81-nV-kZl" secondAttribute="bottom" constant="33" id="DUV-3e-uwW"/>
-                            <constraint firstItem="uVu-PR-0oJ" firstAttribute="top" secondItem="bTG-in-woi" secondAttribute="bottom" constant="13" id="F0h-TW-BRC"/>
+                            <constraint firstItem="uVu-PR-0oJ" firstAttribute="top" secondItem="bTG-in-woi" secondAttribute="bottom" constant="15" id="F0h-TW-BRC"/>
```
erikdstock: same here, just wondering what is going on.

Kiosk/Storyboards/Fulfillment.storyboard
ashfurrow: No problem. This is a somewhat inscrutable XML file that's edited in a WYSIWYG editor within Xcode. I'm not sure about the specific changes to be honest.

ashfurrow:
Correct, the `size` parameter gets mapped into a query string param.

Kiosk/Storyboards/Fulfillment.storyboard
orta: I think it's when what it shows in the Storyboard UI isn't what will be in the app because you have changed constraints but not updated the "positioning" (tends to show as a blue border)

Kiosk/Storyboards/Fulfillment.storyboard
ashfurrow: This is probably because I made the label shorter and moved it down to re-centre it. So the constraint of its bottom edge to the superview space went from 13 to 15 points.

PR #664, opened 2017-05-27
Adds Segment for analytics

### LINKED ISSUE #663: Remove Mixpanel integration (in favor of Segment)
As part of an overarching effort to remove our dependency on (and stop paying for) Mixpanel, Eidelon should stop sending event data directly to Mixpanel via their client library.

This is part of https://github.com/artsy/minotaur/issues/105

ashfurrow:
Fixes #663. New API key is in 1Password vault.


orta:
Pretty sure that's it - yep

PR #665, opened 2017-06-06
Updates project to Xcode 9

ashfurrow:
None of these changes are source breaking, I've checked that the project still compiles in Xcode 8 (which is :100:).

I did briefly try upgrading to Swift 4 (made a few changes I to remove compiler warnings) but they required breaking backwards compatibility with Swift 3, so I think I'll leave things as-is for now. For what its worth, the upgrade process went really smoothly 👌


Kiosk/Sale Artwork Details/SaleArtworkDetailsViewController.swift
```diff
@@ -199,7 +199,7 @@ class SaleArtworkDetailsViewController: UIViewController {
             .rx.observe(NSNumber.self, "highestBidCents")
             .map { observeredCents -> Bool in
                 guard let cents = observeredCents else { return false }
-                return (cents as Int) > 0
+                return (cents as! Int) > 0
```
orta: pedantic: perhaps this would look/feel better with the typecast in the guard as `guard let cents = observeredCents as Int else { return false }`

ashfurrow: Nah, that makes sense. Will do.

PR #666, opened 2017-06-21
Updates CardFlight SDK

ashfurrow:
Version 3.6 fixes a problem with newer card readers, no reason we shouldn't be on the latest and greatest.


orta:
EZ

PR #673, opened 2017-08-28
Adds note about enclosures.

### LINKED ISSUE #671: Add a link to this in the README?
http://vsby.co/work/auction-kiosk

ashfurrow:
Fixes #671.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Project Status
 
 <img src ="https://raw.githubusercontent.com/artsy/eidolon/master/docs/eidolon_preview.jpg">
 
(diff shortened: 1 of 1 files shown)


ashfurrow:
@orta Does Peril have an org-wide ignore list?

orta:
@ashfurrow  - yeah, [here](https://github.com/artsy/artsy-danger/blob/master/spellcheck.json), 

weird bit is that it should actually include READMEs

orta:
oh, it doesn't that's OK, this is fine, will see where it goes

orta:
the dupe is a bit worrying though

PR #675, opened 2017-08-30
Moves to RxOptional and cleans up RxSwift use

### LINKED ISSUE #587: Move to RxOptional
@thanegill took the ideas used with the `Occupiable` protocol and turned them into a solid library: https://cocoapods.org/pods/RxOptional We should consider moving to that library.

ashfurrow:
This fixes #587 by moving our own functions to a pod that already contains them (Eidolon was the inspiration for the pod) and also cleans up our use of RxSwift. Specifically, it removes the custom `doOnNext`, `doOnError`, and `doOnCompleted` functions and replaces their use with the standard `do(onNext:onError:onCompleted:)` operator. Finally, it fixes some intermittently failing snapshot tests (see note left in file).

### DIFF
--- a/KioskTests/ListingsViewControllerTests.swift
+++ b/KioskTests/ListingsViewControllerTests.swift
@@ -19,39 +19,44 @@ class ListingsViewControllerConfiguration: QuickConfiguration {
                 subject.loadViewProgrammatically()
             }
 
+            /*
+            Note that we use a tolerance of 1% to account for antialiasing issues that differ
+            from machine to machine.
+            */
+
             it("grid") {
                 subject.switchView[0]?.sendActions(for: .touchUpInside)
-                expect(subject).to( haveValidSnapshot(usesDrawRect: true) )
+                expect(subject).to( haveValidSnapshot(usesDrawRect: true, tolerance: 0.01) )
             }
             
             it("least bids") {
                 subject.switchView[1]?.sendActions(for: .touchUpInside)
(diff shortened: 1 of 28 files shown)


KioskTests/ListingsViewControllerTests.swift
```diff
@@ -19,39 +19,44 @@ class ListingsViewControllerConfiguration: QuickConfiguration {
                 subject.loadViewProgrammatically()
             }
 
+            /*
+            Note that we use a tolerance of 1% to account for antialiasing issues that differ
+            from machine to machine.
+            */
```
orta: Interesting, the old technique used to be binary bit for bit, I imagine this must be quite a bit slower - but it's not exactly a test suite we run too often 👍

Podfile.lock
```diff
@@ -183,6 +188,6 @@ SPEC CHECKSUMS:
   UIView+BooleanAnimations: a760be9a066036e55f298b7b7350a6cb14cfcd97
   XNGMarkdownParser: aed98c14f0c0eb20064184ddf9bac26c482722b2
 
-PODFILE CHECKSUM: b8b41c512b6c47804dc56fcd2417472abc3cf3f0
+PODFILE CHECKSUM: df9251168a54b50f97a66fb3e48399dfc0a08ab7
 
 COCOAPODS: 1.2.1
```
orta: Update to 1.3.0 - it's considerably faster

PR #669, opened 2017-07-05
WIP: Migrates to Moya 9

ashfurrow:
This PR points our Moya dependency to [this PR](https://github.com/Moya/Moya/pull/1153), a proposal in Moya to move to a more idiomatic way of accessing the Rx extension of Moya. I had some concerns about backwards compatibility and wanted to use Eidolon as a testbed to see how the migrate went.

Things went very smoothly. I had to add the `headers` property to my `TargetType`s incidentally. The actual change for the PR involves the `Networking.swift` file. These changes modify how we customize the Rx extensions in Moya; specifically, our subclass _has_ a provider now (instead of _being_ a provider before). There were a few gotchas with the compiler so hopefully this PR will help someone else.

This is a WIP since Moya 9 isn't officially out yet. 

Tests fail locally but this seems incidental to the Moya changes:

screen shot 2017-07-05 at 1 27 23 pm

PR #677, opened 2017-09-16
Moya 9.0 migration

pedrovereza:
Continues #669.

Eidolon has always been used as the role model app in the Moya community and is constantly used as an example when people are looking for good practices. As a way to thank you folks for the good work, I went ahead and migrated Eidolon to the new major release of Moya :wink:

Some of the changes were done by @ashfurrow in #669, which I rebased against `master` and fixed some conflicts in `Podfile` and `Podfile.lock`. The changelog was also out of sync, but as I wasn't sure about the entries I just took them off for now. That's one thing pending.

I may have missed some *improvements* as I was mostly focused in getting Eidolon to work with Moya 9.0. The [Migration Guide on Moya](https://github.com/Moya/Moya/blob/master/docs/MigrationGuides.md) may have some new things that could be helpful.

Summary of changes:

* `headers` are now specified in the `TargetType` definition;
* `parameters` and `parameterEncoding` were combined into `task`. I was able to remove `parameterEncoding` from all `TargetType`s, but left `parameters` because 1) as there are many endpoints in the API, it feels more readable to compute parameters separately; 2)  [there is a test relying on `parameters`](https://github.com/artsy/eidolon/blob/master/KioskTests/ListingsViewModelTests.swift#L32);
* Couple of functions changed their signature;
* From #669: 
  > Modify how we customize the Rx extensions in Moya; specifically, our subclass has a provider now (instead of being a provider before). 

What's pending:

- [ ] Add changelog entry.

Hope this is helpful :tada:

### DIFF
--- a/Kiosk/App/Networking/ArtsyAPI.swift
+++ b/Kiosk/App/Networking/ArtsyAPI.swift
@@ -54,15 +54,14 @@ extension ArtsyAPI : TargetType, ArtsyAPIType {
         return nil
     }
 
-    public var parameterEncoding: ParameterEncoding {
-        return JSONEncoding.default
+    var task: Task {
+        if let requestParameters = parameters {
+            return .requestParameters(parameters: requestParameters, encoding: JSONEncoding.default)
+        }
+        return .requestPlain
     }
 
-    public var task: Task {
-        return .request
-    }
-
-     var path: String {
+    var path: String {
         switch self {
 
         case .xApp:
@@ -73,7 +72,7 @@ extension ArtsyAPI : TargetType, ArtsyAPIType {
 
         case .auctionInfo(let id):
             return "/api/v1/sale/\(id)"
-            
+
         case .auctions:
             return "/api/v1/sales"
 
@@ -261,12 +260,11 @@ extension ArtsyAuthenticatedAPI: TargetType, ArtsyAPIType {
         return nil
     }
 
-    public var parameterEncoding: ParameterEncoding {
-        return JSONEncoding.default
-    }
-
-    public var task: Task {
-        return .request
+    var task: Task {
+        if let requestParameters = parameters {
+            return .requestParameters(parameters: requestParameters, encoding: JSONEncoding.default)
+        }
+        return .requestPlain
     }
 
     var path: String {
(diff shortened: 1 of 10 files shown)


Kiosk/Bid Fulfillment/Models/BidDetails.swift
```diff
@@ -39,7 +39,7 @@ import Moya
                 // Grab existing endpoint to piggy-back off of any existing configurations being used by the sharedprovider.
                 let endpoint = Networking.endpointsClosure()(target)
 
-                return endpoint.adding(newParameters: ["auction_pin": pin, "number": number, "sale_id": auctionID])
+                return endpoint.adding(newHTTPHeaderFields: ["auction_pin": pin, "number": number, "sale_id": auctionID])
```
ashfurrow: This changed from adding parameters to adding HTTP header fields, which our API doesn't handle in the same way. No biggie – looks like the `adding(newParameters:)` function was removed as part of Moya 9's migration to Tasks, so this was the only suggested function. I only point this out in case folks are wondering how to migrate the use of the old function; my attempt is here: https://github.com/artsy/eidolon/pull/678/files#diff-319cda9244d4aefd8caec7cc37a529c3R42 If anyone has a better way, let me know 👍

Kiosk/App/Networking/ArtsyAPI.swift
```diff
@@ -50,15 +50,18 @@ enum ArtsyAuthenticatedAPI {
...
-        return JSONEncoding.default
+    var headers: [String : String]? {
+        return nil
     }
 
-    public var task: Task {
-        return .request
+    var task: Task {
+        if let requestParameters = parameters {
+            return .requestParameters(parameters: requestParameters, encoding: JSONEncoding.default)
```
ashfurrow: This is one other thing that changed with Moya 9 – looks like a longstanding bug that happened to be fixed that this code relied on. 

Old behaviour: encoding POST requests with the JSON encoder as specified, otherwise encode them in the URL. Pretty easy to work around: https://github.com/artsy/eidolon/pull/678/files#diff-2e89f1c7f10111aa886cbf6eb3922b40R273

PR #678, opened 2017-09-22
Cleanup from RxSwift+Moya upgrades 📱

ashfurrow:
Funny story! There's been a longstanding bug somewhere, which I will try to track down, but this fixes it. Eidolon has been specifying JSON encoding of request parameters for some time, but has been URL encoding them for GET/PUT requests (we only want JSON encoding for POST bodies). Somewhere, probably in Alamofire, the behaviour changed so if you specify JSON encoding, it _always_ encodes a HTTP body with JSON, even for GET requests. This confuses our API, so I've changed it to use JSON only for POST requests. This was a tricky one to track down – using the built-in `NetworkLoggerPlugin` with cURL printing helped a lot.

Additionally, there was a cycle in RxSwift where a Variable was being set inside a subscribe block that executes when the Variable changes. Kind of a weird one, not sure if that's a longstanding bug or something that changed in RxSwift, but the new diagnostic logging helped me track it down 👍

Finally, in the transition to Moya 9, some parameters had [accidentally been added as HTTP header fields](https://github.com/artsy/eidolon/pull/677/files#diff-319cda9244d4aefd8caec7cc37a529c3R42) which broke authenticated requests. Since Moya moved to a new Task infrastructure, augmenting parameters with new ones was a bit tricky.

This is still a WIP because I need to do more testing.

### DIFF
--- a/Kiosk/Bid Fulfillment/RegisterViewController.swift
+++ b/Kiosk/Bid Fulfillment/RegisterViewController.swift
@@ -33,35 +33,20 @@ class RegisterViewController: UIViewController {
         super.viewDidLoad()
 
         coordinator.storyboard = self.storyboard!
-        let registerIndex = coordinator.currentIndex.asObservable()
-        let indexIsConfirmed = registerIndex.map { return ($0 == RegistrationIndex.confirmVC.toInt()) }
+        let indexIsConfirmed = coordinator.currentIndex.map { return ($0 == RegistrationIndex.confirmVC.toInt()) }
 
         indexIsConfirmed
             .not()
             .bindTo(confirmButton.rx_hidden)
             .addDisposableTo(rx_disposeBag)
 
-        registerIndex
+        coordinator.currentIndex
             .bindTo(flowView.highlightedIndex)
             .addDisposableTo(rx_disposeBag)
 
         let details = self.fulfillmentNav().bidDetails
         flowView.details = details
         bidDetailsPreviewView.bidDetails = details
-
-        flowView
-            .highlightedIndex
-            .asObservable()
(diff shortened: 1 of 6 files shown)


ashfurrow:
Cool! I've tested on a device and was able to run through the registration, bidding, and forgot-password flows no problem.

Kiosk/Bid Fulfillment/RegisterViewController.swift
```diff
@@ -33,36 +33,21 @@ class RegisterViewController: UIViewController {
...
-            .subscribe(onNext: { [weak self] (index) in
-                if let _ = self?.fulfillmentNav() {
-                    let registrationIndex = RegistrationIndex.fromInt(index)
-
-                    let nextVC = self?.coordinator.viewControllerForIndex(registrationIndex)
-                    self?.goToViewController(nextVC!)
-                }
-            })
-            .addDisposableTo(rx_disposeBag)
-
```
orta: Cool cool, this threw me off for a second, makes sense

ashfurrow: Yeah, moving to the next VC is handled here instead: https://github.com/artsy/eidolon/pull/678/files#diff-f37ecaa1c56ebb0278e22df551209993R71 So it would get in this loop because calling the `viewControllerForIndex` function in the subscribe block actually _updates_ the index. Like I said, the improved logging in RxSwift helped track this down 👍 Good job @kzaher.

kzaher: Glad it helped :)

PR #683, opened 2018-01-02
Updates Nokogiri to avoid security vulnerability #trivial

ashfurrow:
This updates the Nokogiri dependency, relied upon for Eidolon's deploy script and CI, to avoid a security vulnerability.

Issue #682, opened 2018-01-02
CI is failing


ashfurrow:
Started in https://circleci.com/gh/artsy/eidolon/255 – looks like fonts aren't loading correctly for snapshot tests: https://eidolon-ci.s3.amazonaws.com/254/index.html

ashfurrow:
Probably related: https://github.com/artsy/eigen/issues/2441

PR #684, opened 2018-01-02
Fixes CI

ashfurrow:
This is an attempt to fix CI (see #682). It includes the commit from #683 as well.

This is a WIP because it's only failing on CI. I updated the font pod because that's what usually fixes things, so fingers crossed.
 
EDIT: The fix was just to update the font pod. I also added some Info.plist entries because it's probably good practice (even though Eidolon apparently has worked fine without them to date).


ashfurrow:
Alrighty so I just updated the fonts pod that Eidolon uses from 3.1.1 to 3.1.3, which seems to have fixed the problem (on clean builds I could reproduce it locally too). However, the library update also includes update font files themselves, which subsequently get rendered differently, and broke all the snapshot tests I was trying to fix.

So I updated them. An initial pass using Kaleidoscope didn't reveal any substantive changes, just stuff like this:

screen shot 2018-01-03 at 5 18 11 pm

I'll review inline and un-WIP the PR when I'm satisfied.

PR #706, opened 2019-03-07
Update oauth call to use post

sweir27:
Relates to: https://artsyproduct.atlassian.net/browse/PURCHASE-935


orta:
Hah, yep - you can only make a green/red status on a commit - not a PR. Interesting that it doesn't get overwritten on the 2nd pr though.

PR #708, opened 2019-07-31
[PURCHASE-1292] Fixes manual CC entry

ashfurrow:
This fixes manual CC input. The problem was [when we moved from ReactiveCocoa to RxSwift](https://artsy.github.io/blog/2015/12/08/reactive-cocoa-to-rxswift/) as an FRP framework, I got these bindings in the wrong direction. View model properties were never updated, so the button was always disabled.

This PR fixes those bindings and also requires name entry for manual CC input.

I was getting unit test failures locally, with snapshot screen size differences causing errors with text flowing. I'm curious to see which CI tests fail.


ashfurrow:
Okay, hmm.

So the tests that failed were only the new snapshots (that I didn't add). That means that my failures locally are specific to my machine and not CI, which means it'll be difficult to produce those new snapshots in a way that CI expects. That is to say, something about my local machine is producing consistently incorrect snapshots, so the generated snapshots for the new tests will be also incorrect.

I'll try a reboot, clearing derived data, and resetting the simulator.

ashfurrow:
Ah, classic. The Kiosk uses a different iOS version for unit tests than Eigen, so updating my testing simulator to 11.4 worked 👍

Kiosk/Bid Fulfillment/ManualCreditCardInputViewController.swift
```diff
@@ -36,34 +36,34 @@ class ManualCreditCardInputViewController: UIViewController, RegistrationSubCont
         billingZipWrapperView.isHidden = true
 
         // We show the enter credit card number, then the date switching the views around
-        viewModel
-            .cardFullDigits
+        cardNumberTextField.rx.text
             .asObservable()
-            .bind(to: cardNumberTextField.rx.text)
+            .replaceNilWith("")
```
ashfurrow: `cardNumberTextField.rx.text` produces a string of `Optional<String>`, and `replaceNil("")` will replace the optional values with the empty string.

Kiosk/Bid Fulfillment/ManualCreditCardInputViewController.swift
```diff
@@ -36,34 +36,34 @@ class ManualCreditCardInputViewController: UIViewController, RegistrationSubCont
...
+            .bind(to: viewModel.cardFullDigits)
```
ashfurrow: So here we see where we _were_ binding from `viewModel` -> user interface, but are now binding user interface -> `viewModel`, fixing the bug.

Kiosk/Bid Fulfillment/ManualCreditCardInputViewController.swift
```diff
@@ -77,14 +77,19 @@ class ManualCreditCardInputViewController: UIViewController, RegistrationSubCont
         action
             .errors // Based on errors
             .take(1) // On the first error, then forever
+            .logNext()
```
ashfurrow: I just thought this would be helpful (Auctions Ops/CRT can see the device logs at events).

Kiosk/Bid Fulfillment/ManualCreditCardInputViewController.swift
```diff
@@ -77,14 +77,19 @@ class ManualCreditCardInputViewController: UIViewController, RegistrationSubCont
...
             .startWith(true) // But begin with true
             .bind(to: billingZipErrorLabel.rx_hidden) // show the error label
             .disposed(by: rx.disposeBag)
 
-        viewModel.moveToYear.take(1).subscribe(onNext: { [weak self] _ in
-            self?.expirationYearTextField.becomeFirstResponder()
-        }).disposed(by: rx.disposeBag)
+        viewModel.moveToYear
+            .take(1)
+            .observeOn(MainScheduler.asyncInstance)
```
ashfurrow: This was a weird one – RxSwift gave a really helpful error message here:

https://github.com/ReactiveX/RxSwift/blob/6b2a406b928cc7970874dcaed0ab18e7265e41ef/RxSwift/Rx.swift#L94-L104

Basically, in the `subscribe` closure, we are performing a side-effect that triggers another event to be triggered (importantly, it's triggered before the previous event finishes). `observeOn` is kind of a way to delay the invocation on to the next invocation of the runloop.

PR #709, opened 2019-07-31
Fix keys error

ashfurrow:
I was seeing a local compiler error after updating some keys values, and had to add this `typealias` to get it to compile. I'm curious to see if this will build on CI at all.


ashfurrow:
Yup, failed. That means something's up with my local machine. Closing PR.

PR #711, opened 2019-10-10
[ME-48] Adds checkbox for conditions of sale during registration.

ashfurrow:
I have added a checkbox for agreeing to the Conditions of Sale during registration. I rendered the [Palette checkbox component](https://palette.artsy.net/elements/inputs/checkbox/) into PNGs for both checked states. I've tried to comment why things are done, but let me know what I can clarify.

There are no unit tests for these views, sadly.


Kiosk/Bid Fulfillment/RegistrationConfirmationViewController.swift
```diff
@@ -39,6 +49,23 @@ class RegistrationConfirmationViewController: UIViewController {
...
+        toggleConditionsOfSaleChecked()
+    }
+
+    func toggleConditionsOfSaleChecked() {
+        _conditionsOfSaleChecked.value = !_conditionsOfSaleChecked.value
+
+        if _conditionsOfSaleChecked.value {
+            conditionsOfSaleCheckbox.setImage(UIImage.init(named: "checkmark_checked"), for: .normal)
+        } else {
+            conditionsOfSaleCheckbox.setImage(UIImage.init(named: "checkmark_unchecked"), for: .normal)
```
erikdstock: In react i'd expect the visual result to be declared separately from the effect of toggling the checkbox, but I'm guessing these are more naturally colocated in the context of a view controller.

Kiosk/Storyboards/Fulfillment.storyboard
```diff
@@ -1680,15 +1680,23 @@
...
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
                         <subviews>
+                            <button opaque="NO" contentMode="left" fixedFrame="YES" contentHorizontalAlignment="leading" contentVerticalAlignment="center" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="R87-pJ-OQN">
+                                <rect key="frame" x="0.0" y="86" width="44" height="44"/>
+                                <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMaxY="YES"/>
+                                <state key="normal" image="checkmark_unchecked"/>
+                                <connections>
+                                    <action selector="conditionsOfSaleCheckmarkTapped:" destination="q7z-ij-DNu" eventType="touchUpInside" id="gPa-dm-9Dt"/>
+                                </connections>
+                            </button>
```
erikdstock: Was this done via the storyboard editor?

Kiosk/Bid Fulfillment/RegistrationConfirmationViewController.swift
```diff
@@ -1,10 +1,17 @@
...
 let targetConditionsOfSaleText = "Conditions of Sale"
 
 class RegistrationConfirmationViewController: UIViewController {
     @IBOutlet weak var confirmationLabel: ARSerifLabel!
+    @IBOutlet weak var conditionsOfSaleCheckbox: UIButton!
+
+    fileprivate let _conditionsOfSaleChecked: Variable<Bool> = Variable(false)
+    var conditionsOfSaleChecked: Observable<Bool> {
+        return _conditionsOfSaleChecked.asObservable()
+    }
```
erikdstock: Why private + getter, and why observable? I think I understand:
- the boolean itself is private and not so important
- the button in the storyboard file down below has an action/attribute to toggle the value (`<action selector="conditionsOfSaleCheckmarkTapped:" destination="q7z-ij-DNu" eventType="touchUpInside" id="gPa-dm-9Dt"/>`)
- somewhere up in `RegisterViewController` the observable gets `conditionsOfSaleChecked.bind(to: self.confirmButton.rx.isEnabled)` which means that if it is checked the button gets enabled.

Kiosk/Bid Fulfillment/RegistrationConfirmationViewController.swift
ashfurrow: Yeah exactly.

Kiosk/Storyboards/Fulfillment.storyboard
ashfurrow: That's right – I didn't even look at the text the diff until I opened the PR.

Kiosk/Bid Fulfillment/RegistrationConfirmationViewController.swift
ashfurrow: Great question! The critical thing to note is that `Variable` is read/write, while `Observable` is read-only. By declaring the read/write version as `file private`, we make it so only this view controller can set the value. But by declaring a read-only version that others can access, the other controller can bind as you described.

ashfurrow:
Thanks @erikdstock! Let me know what else I can clarify before we merge. I've got one more Kiosk ticket to get to, so I'll probably ping you again for a review 👍

erikdstock:
All good, merge away.

PR #712, opened 2019-10-18
[ME-47] Phone number plus signs

ashfurrow:
There are a few changes here related to international phone numbers:

- Allow `+` signs to be entered into the arbitrary text field, using the OS keyboard.
- Add `+` sign to keypads that are used for phone numbers; these are custom UI for the user to enter phone numbers.

Here's what it looks like:

There was a conflict in how phone numbers were getting entered, and how they were getting formatted. There was no easy all-purpose formatter that takes into account our needs, so I confirmed with Melanie that it would be okay to just not format the phone numbers as the user entered them. So we used to format `5555555555` as `"555-555-5555"`, but that's not possible with international phone numbers without a lot of work. Now, we'll show exactly what the user inputs.

Oh, and I got to learn a little bit about custom Interface Builder files. This lets us specify if a keypad is for phone numbers or for bid values in IB, which looks like this:

Screen Shot 2019-10-18 at 15 43 39

It will actually live-update according to the selection, which is something Orta tried back in 2014 but IB was still too buggy. Now it works 🎉


dleve123:
@ashfurrow The gifs are really fast, so a bit hard to tell, but are we now _requiring_ that a user enter a `+` before a phone number in this app, or just permitting it? Also, Will this PR change how any phone number data is persisted in gravity?

ashfurrow:
@dleve123 We aren't requiring the user to enter a `+` sign, it's totally optional. Gravity persists whatever arbitrary string we send over when we [create the new user account](https://github.com/artsy/gravity/blob/4c45fd69a1c9df8f87936fcee94fbc567bf21b8b/app/api/v1/users_endpoint.rb#L16-L17) so there's no back-end change needed.

dleve123:
Yeah, was just playing around with Gravity and Force and it appears that we don't perform any phone number standardization and the validation on the backend is incredibly minimal. For example, entering a phone number of `dd` will _fail_ validation, entering a phone number of `917d` will _pass_ it.

This seems pretty problematic actually for our larger "buyer vetting" strategy.

ashfurrow:
I've only done some cursory exploration of how we use+validate phone numbers, and I agree there's a lot of room for improvement there 👍

Kiosk/Bid Fulfillment/KeypadContainerView.swift
```diff
@@ -4,39 +4,58 @@ import RxSwift
...
 
-        let bundle = Bundle(for: type(of: self))
-        let image  = UIImage(named: "KeypadViewPreviewIB", in: bundle, compatibleWith: self.traitCollection)
-        let imageView = UIImageView(frame: self.bounds)
-        imageView.image = image
+        if self.isPhoneNumberEntry {
+            keypad.setLeftButton(.plus)
+        } else {
+            keypad.setLeftButton(.delete)
+        }
```
erikdstock: Is there a delete/clear button for non-phone numbers?

ashfurrow: That's right, we still use this UI for entering bid amounts.

PR #718, opened 2020-04-20
change public spec source to CDN

muhasturk:
You should use CDN for public pod spec after cocoapods 1.7.2


ashfurrow:
Hey, thanks for the PR 👍 I've pulled the commit into #719 and did the other cleanup work required to get it ready (it is unfortunately not enough to just change the `source` in `Podfile`). Take care!

PR #724, opened 2020-12-03
fix: handles nil fonts correctly for OSS contributors #trivial

ashfurrow:
An open source contributor [brought up a problem](https://github.com/ashfurrow/xcode-hardware-performance/pull/137#issue-527887411) where the OSS font-loading wasn't working (ie: the font is returning `nil`) and our code does not expect that. I tried debugging this but the iOS font pod closed-vs-open fonts are notoriously hard for Artsy staff to debug because of how we determine if someone "should" have the closed fonts ([see code here](https://github.com/artsy/Artsy-OSSUIFonts/blob/master/Pod/Scripts/ArtsySetup.rb)). So instead, I looked for places where we were force-unwrapping these fonts and added code to make sure that we handle the `nil`s correctly.

CI might not pass on this but the PR should be safe to merge – I tested locally.

/cc @jlnr @ondrejhanak

### DIFF
--- a/Kiosk/App/Models/Artwork.swift
+++ b/Kiosk/App/Models/Artwork.swift
@@ -94,12 +94,12 @@ import SwiftyJSON
 private func titleAndDateAttributedString(_ title: String, dateString: String) -> NSAttributedString {
     let workTitle = title.isEmpty ? "Untitled" : title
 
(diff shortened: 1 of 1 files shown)


ashfurrow:
#squashongreen

